PR #10, opened 2017-05-17
BUG: data.maybe_open couldn't use open file handle

andyfaff:


pkienzle:
Again, didn't see the notification 'til now.  Thanks!

PR #11, opened 2017-05-18
ENH: de candidates optimization

andyfaff:
This PR speeds the candidate selection up by a factor of 3 (from 66us to 20us for a (100, 20) population shape). It's also more pythonic.
The original `_candidates` function also had a bug. If `exclude is None` then the `selection[selection>=exclude]` line doesn't work:

    TypeError: '>=' not supported between instances of 'int' and 'NoneType'

### DIFF
--- a/bumps/mystic/optimizer/de.py
+++ b/bumps/mystic/optimizer/de.py
@@ -141,9 +141,12 @@ def _candidates(pop, n, exclude=None):
     Select *n* random candidates from *pop*, not including the
     candidate at index *exclude*.
     """
-    selection = choose_without_replacement(len(pop)-1, n)
-    selection[selection>=exclude] += 1
-    return pop[selection]
+    idxs = list(range(len(pop)))
+    if exclude is not None:
+        idxs.pop(exclude)
+    np.random.shuffle(idxs)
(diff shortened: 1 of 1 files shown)


pkienzle:
Sorry... I'm not getting notifications from bumps, so I missed this PR. Don't know why.

I don't see how your code is any more pythonic.  I prefer the more compact variation:
{{{
    selection = np.random.choice(len(pop)-1, size=k, replace=False)
    return pop[selection + (selection >= exclude)] if exclude is not None else pop[selection]
}}}

That said, random.shuffle() seems to be faster than random.choice(), even for large n, small k.  This is despite random.shuffle() being O(n=num available) and random.choice() being O(k=num chosen).  So I will follow your suggestion and use shuffle.  

I've twiddled it a bit to avoid making lists and turning them into arrays so I'm actually using the following:
{{{
def _candidates(pop, k, exclude=None):
    n = len(pop)
    if exclude is not None:
        selection = np.arange(n-1, dtype='i')
        if exclude < n-1:
            selection[exclude] = n-1
    else:
        selection = np.arange(n, dtype='i')
    np.random.shuffle(selection)
    return pop[selection[:k]]
}}}

Thanks!

PR #12, opened 2017-12-06
plot_vars - set the figure size according to the number of plots

e-rus:
I edited the code for the vars figure to scale the figure size according to the number subplots.  The margins and the spacing between the plots remains the same regardless of how many plots there are, so the axes will remain readable even for models with a large number of parameters.

### DIFF
--- a/bumps/dream/views.py
+++ b/bumps/dream/views.py
@@ -174,7 +174,7 @@ def tile_axes(n):
 
 def plot_var(draw, vstats, var, cbar, nbins=30):
     values = draw.points[:, var].flatten()
-    _make_logp_histogram(values, draw.logp, nbins, vstats.p95,
+    _make_logp_histogram(values, draw.logp, nbins, vstats.p95_range,
                          draw.weights, cbar)
     _decorate_histogram(vstats)
 
@@ -182,16 +182,21 @@ def plot_var(draw, vstats, var, cbar, nbins=30):
 def _decorate_histogram(vstats):
     from matplotlib import pyplot as plt
     from matplotlib.transforms import blended_transform_factory as blend
+
+    l95, h95 = vstats.p95_range
+    l68, h68 = vstats.p68_range
+
     # Shade things inside 1-sigma
+<<<<<<< HEAD
     plt.axvspan(vstats.p68[0], vstats.p68[1],
                   color='gold', alpha=0.5, zorder=-1,
                 ec='none')
+=======
+    pylab.axvspan(l68, h68, color='gold', alpha=0.5, zorder=-1)
+>>>>>>> upstream/master
     # build transform with x=data, y=axes(0,1)
     axi = plt.gca()
     transform = blend(axi.transData, axi.transAxes)
-
-    l95, h95 = vstats.p95
-    l68, h68 = vstats.p68
 
     def marker(symbol, position):
         if position < l95:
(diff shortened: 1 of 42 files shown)


bumps/dream/views.py
```diff
@@ -61,47 +61,116 @@ def plot_all(state, portion=1.0, figfile=None):
...
+                break
+            dims = [l_margin_f+i*(tile_W_f+h_space_f),
+                    top-j*(tile_H_f+v_space_f),
+                    tile_W_f,
+                    tile_H_f]
+            ax.append(fig.add_axes(dims))
+            plt.sca(ax[k])
+            plot_var(draw, all_vstats[k], k, cbar)
+            ax[k].set_facecolor('none')
+            k+=1
```
pkienzle, line 160: Done.

bumps/dream/views.py
```diff
@@ -61,47 +61,116 @@ def plot_all(state, portion=1.0, figfile=None):
...
-    nh = int(math.floor(math.sqrt(n*phi*figheight/figwidth)))
-    if nh < 1:
-        nh = 1
-    nw = int(math.ceil(n/nh))
-    return nw, nh
+    from numpy import ceil, sqrt
+    
+    cols = int(ceil(sqrt(n)))
+    rows = int(ceil(n/float(cols)))
+    return cols, rows
```
pkienzle, line 173: Done.

bumps/dream/views.py
```diff
@@ -112,17 +181,17 @@ def plot_var(draw, vstats, var, cbar, nbins=30):
...
-    import pylab
+    from matplotlib import pyplot as plt
     from matplotlib.transforms import blended_transform_factory as blend
 
     l95, h95 = vstats.p95_range
     l68, h68 = vstats.p68_range
 
     # Shade things inside 1-sigma
-    pylab.axvspan(l68, h68, color='gold', alpha=0.5, zorder=-1)
+    plt.axvspan(l68, h68,color='gold', alpha=0.5, zorder=-1, ec='none')
```
pkienzle, line 191: The `ec='none'` didn't get included in master. I have no opinion on whether the 1-σ region should have a box drawn around it.

bumps/dream/views.py
```diff
@@ -187,11 +256,17 @@ def __init__(self, low, high):
...
 
-    ticks = (vmin, vmax)
+    ticks = []#(vmin, vmax)
     formatter = MinDigitsFormatter(vmin, vmax)
-    cb = mpl.colorbar.ColorbarBase(ax, cmap=cmap, norm=norm,
+    cb = mpl.colorbar.ColorbarBase(axi, cmap=cmap, norm=norm,
                                    ticks=ticks, format=formatter,
-                                   orientation='horizontal')
+                                   orientation='vertical',
+                                   ticklocation='right')
```
pkienzle, line 264: `ticklocation='right'` is not included in the current code. Should it be?

bumps/errplot.py
```diff
@@ -75,7 +75,7 @@ def reload_errors(model, store, nshown=50, random=True):
                                   nshown=nshown, random=random)
 
 
-def calc_errors_from_state(problem, state, nshown=50, random=True):
+def calc_errors_from_state(problem, state, nshown=50, random=True, selection={}):
```
pkienzle, line 78: Not related to the variable plot figure size, so should be done as a separate PR. 

Should use `selection=None` for no selection. Make sure `state.sample` accepts `selection=None` and the code below won't be needed.

pkienzle (commented):
The main portion of this PR (setting figure size based on the number of plots) has already been incorporated into master. There are a couple of details that you may want to tweak as noted in the detailed comments. Some of the changes belong in separate PRs; these are also noted in the comments.

pkienzle:
I'm closing this PR since the main change has already been included in master. Feel free to submit other changes as separate PRs or turn them into issues.

PR #14, opened 2017-12-26
Allow for numerical precision in test

llimeht:
On x86 hardware (Python 2.7.14 on Debian's i386 port) the test against a known problem fails due to numerical precision. The attached patch uses `numpy`'s `assert_almost_equal` to test more robustly; one could reinvent an `assert abs(a - b) / a < delta` instead but since `numpy` is already used throughout, there's no harm in using it here too.

```Check that results are correct for a known problem. ... > /build/python-bumps-FiUVML/python-bumps-0.7.6/.pybuild/pythonX.Y_2.7/build/bumps/wsolve.py(447)test()
-> assert py == poly(px), "direct call to poly function fails"
(Pdb) py
array([ 13.18486645])
(Pdb) px
array([ 1.5])
(Pdb) poly(px)
array([ 13.18486645])
(Pdb) py - poly(px)
array([  1.77635684e-15])
```

### DIFF
--- a/bumps/wsolve.py
+++ b/bumps/wsolve.py
@@ -422,6 +422,7 @@ def test():
     """
     Check that results are correct for a known problem.
     """
+    from numpy.testing import assert_array_almost_equal_nulp
     x = np.array([0, 1, 2, 3, 4], 'd')
     y = np.array([2.5, 7.9, 13.9, 21.1, 44.4], 'd')
     dy = np.array([1.7, 2.4, 3.6, 4.8, 6.2], 'd')
@@ -446,7 +447,7 @@ def test():
     assert dperr < 1e-14, "||dp-Tdp||=%g" % dperr
     assert cierr < 1e-14, "||ci-Tci||=%g" % cierr
     assert pierr < 1e-14, "||pi-Tpi||=%g" % pierr
-    assert py == poly(px), "direct call to poly function fails"
+    assert_array_almost_equal_nulp(py, poly(px), nulp=8)
 
 if __name__ == "__main__":
     #test()


llimeht:
Do I understand your explanation of why there might be differences as saying that it's not a matter of limited numerical precision in floats but of fundamentally different algorithms giving different values? My reading of `assert_array_almost_equal_nulp` is that it is for comparisons where the implementation details of the float storage is a problem, while relative tolerance tests (such as `assert_allclose` or `assert_almost_equal`) are for comparisons where it is not expected that the results would be identical if it were not for finite precision floats being used.

(Happy either way, just want to pick the right function for the job.)

pkienzle:
Short answer: in this case assert_array_almost_equal_nulp would be the better choice, with nulp=8 allowing the last three bits to differ.  Or even 4.

Looking a bit closer, this is a linear system, so effectively it is the same algorithm, though one code path goes through BLAS and the other is presumably cython (haven't looked at the polyval implementation).  Depending on exact register usage and processor (Intel processors keep a guard bit on numbers retained in registers that can be discarded if the value is written to memory). Regardless, since the test has been passing on my machines using floating point equality (linux, windows and mac but all x86), that's a pretty good hint that the values should be the same except for the last bit or so.

The all_close test is using 7 digits, which is suitable for single precision, but is way too loose for this case.  There are some cases where that much precision is lost compared to an analytic solution (e.g., optimizers may stop when df is less than 1e-7), but not here.  In general, I will adjust a test to use as many bits as I can get away with, sometimes adjusting the algorithm to make the results more precise.

llimeht:
Updated to use `assert_array_almost_equal_nulp` with `nulp=8`.

PR #17, opened 2018-05-08
Use numpy legacy output format when doing doctests

llimeht:
numpy 1.14 changes the output format for printing and would break the doctests throughout; tell numpy to use the 1.13 format instead.

(This PR is built on top of the travis-ci one so that it is tested; I can rebase if needed so that it's only the single commit.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,10 @@
+language: python
+
+python:
+  - "2.7"
+  - "3.6"
+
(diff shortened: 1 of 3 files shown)


pkienzle:
Probably too soon to require numpy>=1.14 so we are stuck with legacy cruft for now even though it is ugly.  How unfortunate...

PR #34, opened 2020-08-20
FIX: simplify handling of figure queue for embedded matplotlib > 3.3.0

### LINKED ISSUE #33: EmbeddedPylab breaks on matplotlib 3.3.1 (latest)
There was an update to the _pylab_helpers in matplotlib between 3.3.0 and 3.3.1, and the EmbeddedPylab utility in bumps breaks after this change (they removed _activeQue)
see https://github.com/matplotlib/matplotlib/commit/7fcf123051c0b8402f779b68ac9988a2369ed329

pkienzle:
Fixes #33

### DIFF
--- a/bumps/gui/util.py
+++ b/bumps/gui/util.py
@@ -64,16 +64,21 @@ class EmbeddedPylab(object):
         # delay loading pylab until matplotlib.use() is called
         import pylab
         from matplotlib._pylab_helpers import Gcf
+        # Note: don't need to track and restore the current active since it
+        # will automatically be restored when we pop the current figure.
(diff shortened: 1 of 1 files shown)


pkienzle:
If you open up the python console in bumps/refl1d and plot something I bet it'll go to the last plot generated rather than opening a new figure.

pkienzle:
...now I have a concern: I wonder what figure number is being assigned to our embedded figures?  Is it likely to collide with a figure number when someone creates a plot window from the console? Can we assign it a number that'll never be used, like -1?

PR #40, opened 2020-11-11
Drop Python 2.7 support from tests

llimeht:
The package uses the matrix multiplication infix operator which is Python 3.5
or later and that makes the most recent 0.7.17 release Python 3.5+. I guess that makes it time to stop testing against Python 2.7.

(There is more that can be done to drop old code, but this is the first step)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,6 @@
 language: python
 
 python:
-  - "2.7"
   - "3.6"
 
 install:


pkienzle:
Alas, not quite yet. We're still waiting to see if SasView releasing 4.3 on python 2.7.
 
I cut a new release. Worst part? The offending code is not on the release code path, and was only used for checking the correctness of the algorithm.

PR #46, opened 2021-01-13
Edited state.py to compress mc files with lzma. Also added ability to…

e-rus:
… read lzma compressed mc files. It seems like lzma can give significantly higher compression ratios than gzip.

### DIFF
--- a/bumps/dream/state.py
+++ b/bumps/dream/state.py
@@ -84,10 +84,15 @@ from .outliers import identify_outliers
 from .util import draw, rng
 from .gelman import gelman
 
-EXT = ".mc.gz"
-CREATE = gzip.open
+import lzma
+from io import StringIO, BytesIO
+
+#EXT = ".mc.gz"
+#CREATE = gzip.open
 #EXT = ".mc"
 #CREATE = open
+EXT = ".mc.xz"
+CREATE = open
 
 # CRUFT: python 2 uses bytes rather than unicode for strings
 try:
@@ -99,6 +104,14 @@ except NameError:
     # python 3.x
     def write(fid, s):
         fid.write(s.encode('utf-8') if isinstance(s, str) else s)
+
+def savetxtlzma(fid, ame, a):
+    savetxt(fid, a)
+    f = open(ame, 'wb')
+    s = lzma.compress(fid.getvalue())
+    f.write(s)
+    f.close()
+    fid.close()
 
 class NoTrace:
     def write(self, data):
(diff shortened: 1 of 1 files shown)


bmaranville:
lzma is also crazy fast

pkienzle:
(1) What kind of compression ratios are you seeing?

If you get good mixing in the fit then I don't expect significant compression; maybe 2.5x (20 bytes for a double string vs. 8 bytes for a double binary). That's what I get for both algorithms when I feed it gaussian random numbers.

Slightly more realistic, if I share the bunch of digits for the different variables I get between 3x and 4x compression with both gzip and lzma, with lzma about 15% smaller. I haven't tested on a stuck fit with lots of repeat samples.

I'm also finding that lzma takes 23s to write 1M samples, compared to 3.5s for gzip and 0.6s for plain write. I'm not sure this is a worthwhile tradeoff.

I tested using:
```python
import gzip, lzma
from numpy.random import rand, randn

def tic():
    from time import perf_counter
    global _tic_time
    _tic_time = perf_counter()

def toc(echo=True):
    from time import perf_counter
    delta = perf_counter() - _tic_time
    if echo:
        print(f"Elapsed time is {delta:g} seconds")
    return None if echo else delta

# 1M samples with 10 repeats of 10 variables with 5-6 digits in common for each variable
data = randn(100,10000) + 1000000*np.array(rand(10).tolist()*10)[:, None]
fid=open('/tmp/pure', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()
fid=gzip.open('/tmp/puregz', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()
fid=lzma.open('/tmp/purelz', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()
!ls -l /tmp/pure*
```

(2) The test is failing on python 2.7. This is because it is missing lzma. We either need to install the backport for py2.7 or drop py2.7 support.

Maybe the following in .github/workflows/test.yml, after the python dependencies:

```yaml
    - name: Install Python 2.7 dependencies
      if: matrix.cfg.py == 2.7
      run: |
        python -m pip install backport.lzma
```

You would also need to update the code with:

```python
try:
    import lzma
except ImportError: # CRUFT: py2.7 support for lzma (requires pip install backports.lzma)
    from backports import lzma
```

(3) I presume there are command line lzma tools in case you want to access the mc files from something other than python?

pkienzle:
(4) No need to buffer the compressed data through BytesIO. You can use the same `CREATE` magic that I used for gzip. It doesn't seem to change either compression ratio or run time to do so, but potentially uses a lot less memory.

bmaranville:
import gzip, lzma, lz4.frame
from numpy.random import rand, randn
from numpy import savetxt
import numpy as np

def tic():
    from time import perf_counter
    global _tic_time
    _tic_time = perf_counter()

def toc(echo=True):
    from time import perf_counter
    delta = perf_counter() - _tic_time
    if echo:
        print(f"Elapsed time is {delta:g} seconds")
    return None if echo else delta

# 1M samples with 10 repeats of 10 variables with 5-6 digits in common for each variable
data = randn(100,10000) + 1000000*np.array(rand(10).tolist()*10)[:, None]
fid=open('/tmp/pure', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()
fid=gzip.open('/tmp/puregz', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()
fid=lzma.open('/tmp/purelz', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()
fid=lz4.frame.open('/tmp/purelz4', 'wb'); tic(); savetxt(fid, data); toc(); fid.close()

Elapsed time is 0.713269 seconds
Elapsed time is 3.80463 seconds
Elapsed time is 23.4334 seconds
Elapsed time is 0.776627 seconds


________________________________

bmaranville:
-rw-r--r--   1 brian brian  24M Feb 16 14:58  pure
-rw-r--r--   1 brian brian 7.7M Feb 16 14:58  puregz
-rw-r--r--   1 brian brian 6.5M Feb 16 14:59  purelz
-rw-r--r--   1 brian brian  14M Feb 16 14:59  purelz4
________________________________

pkienzle:
Superseded by hdf5 session file outputs.

PR #49, opened 2021-02-19
use function imports from operator instead of eval

bmaranville:


pkienzle (approved):
Very nice! I'm happy to see one more exec gone!

You may want to check that operator is in a closed set of valid operators. This reduces the attack surface slightly should we ever get to the point where we have a fully declarative model definition that we are loading from across the net. Maybe some combination of `attrgetter` (the '.' operator) can retrieve the `file` function from `__builtins__` and cause mischief.

Sadly there are no automated tests for parameters. Here's something minimal you can drop in the bottom of `parameter.py`:
```python
def test_operator():
    a = Parameter(1, name='a')
    b = Parameter(2, name='b')
    a_b = a + b
    a.value = 3
    assert a_b.value == 5.
    assert a_b.name == '(a + b)'
```

PR #50, opened 2021-02-19
Apply parameters gui

bmaranville:
Add menu option to apply parameters file to current model

### DIFF
--- a/bumps/gui/app_panel.py
+++ b/bumps/gui/app_panel.py
@@ -36,7 +36,7 @@ import wx
 import wx.aui
 
 from .. import plugin
-from ..cli import load_model
+from ..cli import load_model, load_best
 from ..util import redirect_console
 from ..dream import stats as dream_stats
 
@@ -59,6 +59,7 @@ PYTHON_FILES = "Script files (*.py)|*.py"
 DATA_FILES = "Data files (*.dat)|*.dat"
 TEXT_FILES = "Text files (*.txt)|*.txt"
 ALL_FILES = "All files (*.*)|*"
+PARS_FILES = "Parameter files (*.pars)|*.pars"
 
 # Custom colors.
 WINDOW_BKGD_COLOUR = "#ECE9D8"
@@ -153,6 +154,10 @@ class AppPanel(wx.Panel):
                                   "Open existing model")
         frame.Bind(wx.EVT_MENU, self.OnFileOpen, _item)
         #file_menu.Enable(id=wx.ID_OPEN, enable=False)
+        _item = file_menu.Prepend(wx.ID_ANY,
+                                  "&Apply Pars File",
+                                  "Apply parameters from .pars file to loaded model")
+        frame.Bind(wx.EVT_MENU, self.OnParsFileOpen, _item)
         _item = file_menu.Prepend(wx.ID_NEW,
                                   "&New",
                                   "Create new model")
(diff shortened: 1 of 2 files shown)


bmaranville:
Patrick says he tested this and it seemed to work.

bumps/gui/app_panel.py
```diff
@@ -585,6 +609,10 @@ def load_model(self, path):
         self._reload_path = path
         model = load_model(path)
         signal.model_new(model=model)
+    
+    def apply_parameters(self, path):
+        load_best(self.model, path)
```
pkienzle, line 614: `load_best` goes to some effort to match the parameter names in the current model, ignoring the new ones and leaving old ones unchanged. I believe this behaviour will be the same in this code.

During this process, `load_best` sets an "undefined" vector for the model so that it knows which parameters are not included in the loaded .par file. Within `init_pop` the undefined parameters will be initialized across their range using LHS irrespective of the population initializer set in the options.

The undefined parameters will stay randomized for all subsequent fits in the GUI until another parameter set is reloaded. This is probably not what the user wants. I think the simplest short-term fix is to remove `undefined` from the problem in `init_pop.generate`. At least then it will only affect the current fit, not all subsequent fits.

The idea is that you can randomize the parameter for the next round of fitting by commenting it out in the .par file. This was a horrible hack for a feature that is not advertised and therefore not used, except maybe by Joe if he remembers that it exists. Instead there should be explicit control of the randomized parameters on the command line and in the GUI. I added a ticket for this.

bumps/gui/app_panel.py
```diff
@@ -362,6 +367,25 @@ def OnFileOpen(self, event):
...
+        dlg = wx.FileDialog(self,
+                            message="Select Parameters File",
+                            #defaultDir=os.getcwd(),
+                            #defaultFile="",
+                            wildcard=(PARS_FILES),
+                            style=wx.FD_OPEN|wx.FD_CHANGE_DIR)
+
+        # Wait for user to close the dialog.
+        status = dlg.ShowModal()
+        path = dlg.GetPath()
```
pkienzle, line 382: If the par file has the same name as the model file then selecting the directory is good enough. I can't tell from the wx docs whether the dialog will allow you to select a directory. I'm not sure it matters either way, but we should check what happens. If the user chooses a directory but there is no corresponding par file then a warning should be given.

bmaranville:
This was intended just to provide a way for users to load a model and then apply parameters, the same as they do from the command line.  I don't know all of the other implications.

pkienzle:
The patch is okay, I just want a couple of tweaks.

(1) Remove `wx.FD_CHANGE_DIR` and make sure it still works.

I'm not sure that this is necessary, but not changing the directory is safer. It may affect where files are stored, though I haven't checked.

(2) Replace your `apply_parameters` with:
```python
    def apply_parameters(self, path):
        load_best(self.model, path)
        # TODO: remove this once LHS randomization is controlled from the command line. Ticket #51
        if hasattr(problem, 'undefined'):
            del problem.undefined
        signal.update_parameters(model=self.model)
```
This protects against weirdness when you load parameters from a simpler model (so `undefined` is set), then run a bunch of L-M fits (which don't touch `undefined`), then switch to Dream. Here the LHS randomization will kick in and the user will be confused.

(3) Add `del problem.undefined` to [this](https://github.com/bumps/bumps/blob/1d9622d4b633d33625e03ee15c9bf6f2320f0d9d/bumps/initpop.py#L77-L80) if-block

This comes into play when the user adds parameters to the model then reruns bumps with the old .par file. Until 'undefined' is reset the new parameters will be initialized with LHS for every DREAM run.

pkienzle (approved):
Code looks fine. Feel free to merge if the GUI works with DREAM.

bmaranville:
P. Quarterman reports that it works in the GUI, with all the new changes.

PR #56, opened 2021-03-25
restructure parameter and expression handling

pkienzle:


bmaranville:
I'll probably restructure again so that UnaryExpression, BinaryExpression and VariadicExpression (or just Expression) are separate, otherwise it's impossible to validate the number of arguments from the schema (what does sin(a,b,c) mean?)

pkienzle:
I'm not sure you need to count arguments.

Assuming you want/need to do so, perhaps leave `Expression` as the top-level container but have a callable attribute inside which is an (operator, args) pair. Make operator an enum specific to the particular type signature and arguments a tuple of those types.

For example, arctan2:
```python
@schema()
class Expression:
    fn: Union[Callable_f_f, Callable_ff_f]
    ...

@schema()
class Callable_ff_f:
    op: Operators_ff_f
    args: Tuple[FloatArgument,FloatArgument]

FloatArgument = Union[float,ValueType]
class Operators_ff_f[Enum]:
    arctan2 = "arctan2"
    ...

```

This can then be extended to additional type signatures, for example if we wanted to include a function in the suite of available functions that accepts a string rather than a float.

bmaranville:
While it could be left to the user to make sure that the number of arguments won't break the model, isn't this kind of the point of a schema (to validate the inputs?)

bmaranville:
I like the idea of putting it off for now though - I think you're right that it doesn't matter very much.

pkienzle:
... and we could make `Constant` into a unary function by adding `identity = lambda x: x` to the available functions, whichever seems simpler to you.

Note that values returned by parameters() methods in this PR uses ValueType at the base, not Parameter:
```python
    ParameterStructure = Union[Dict[str, Parameters], List[Parameters], Tuple[Parameters]?, ValueType, None]
```
That is, by using `Constant` in the model we have an object without an `equals()` method and a value property without a setter, so we can't override it with our own value. Of course, you can replace the constant with a parameter or expression but the model will still ignore it, so no effect.

Perhaps the better option is to remove `Constant` entirely? It is only used in refl1d.QProbe so that there are slots for theta offset and sample broadening (but see [this comment](https://github.com/reflectometry/refl1d/issues/47#issuecomment-808361692)).

PR #57, opened 2021-04-01
minimal example of model constraints

pkienzle:


bmaranville:
There is an inconsistency in the Fitness class... in the constraints_nllf method it checks if self.constraints is callable, but it does not in the constraints_satisfied method.

bmaranville:
Some of these changes appear to be bugfixes for code already in the dataclass_overlay branch - i will copy over the ones I can find.

pkienzle:
I pulled beforehand to make sure that Ι had the latest version. Is it possible you didn't push some updates?

constraints_nllf and constraints_satisfied should not both exist.  There's no reason to evaluate all the constraints twice. I think constraints_nllf is more flexible since it allows us to use soft constraints later if we so choose whereas constraints_satisfied enforces hard constraints. If we go with pure constraints-as-prior and eliminate the penalty function I think we might be okay within the optimizers. For the simple tests I've thrown at them they seem to work, but I didn't test systematically.

We do want to make sure the code short-circuits and doesn't compute the model when the constraints fail (i.e., constraints_nllf is infinite).

bmaranville:
No - I am copying bug fixes *from* your PR to the other branch.  You fixed some things unrelated to the model-specific constraints in your branch.

pkienzle:
That's fine.  This branch can be rebased when you are done.

PR #59, opened 2021-04-27
Refactor serialization

pkienzle:
Formalize the schema for bumps fits and provide serialization/deserialization from JSON.


bumps/fitproblem.py
```diff
@@ -172,136 +190,117 @@ def FitProblem(*args, **kw):
...
-                self.name = 'FitProblem'
+            self._constraints_function = lambda: 0.
+            # TODO: do we want to allow "constraints=a<b" or do we require a sequence "constraints=[a<b]"?
+            self.constraints = constraints if constraints is not None else []
+        if freevars is None:
+            names = ["M%d" % i for i, _ in enumerate(models)]
+            freevars = parameter.FreeVariables(names=names)
+        self.freevars = freevars
+        self._models = models
+        self.__class__.models = property(self._get_models)
```
pkienzle, line 243: Shouldn't have to diddle class method during object creation. What happens when the next object is created? Why can't models be made a property in the usual way?

Deprecate the use of 'problem.models` to iterate over models since its semantics are surprising. Instead provide a method `cycle_models()` or some such to indicate that `models` is not an ordinary list, but instead is only valid for one model at a time due to free variable substitution.

bumps/fitproblem.py
```diff
@@ -172,136 +190,117 @@ def FitProblem(*args, **kw):
...
-        """
-        Parameters associated with the model.
-        """
-        return self.fitness.parameters()
+        """Return parameters from all models"""
+        pars = {'models': [f.parameters() for f in self.models]}
+        free = self.freevars.parameters()
+        if free:
+            pars['freevars'] = free
+        return pars
```
pkienzle, line 284: Shouldn't this be `def parameters(self):` for consistency?

bumps/fitproblem.py
```diff
@@ -172,136 +190,117 @@ def FitProblem(*args, **kw):
...
-            'fitness': to_dict(self.fitness),
-            'partial': self.partial,
-            'soft_limit': self.soft_limit,
+            'models': to_dict(self._models),
+            'weights': self.weights,
             'penalty_nllf': self.penalty_nllf,
             # TODO: constraints may be a function.
             'constraints': to_dict(self.constraints),
+            'freevars': to_dict(self.freevars),
         }
```
pkienzle, line 296: replace `to_dict` with `__getstate__`, `__setstate__`.

bumps/fitproblem.py
```diff
@@ -336,8 +335,8 @@ def getp(self):
         return np.array([p.value for p in self._parameters], 'd')
 
     def bounds(self):
-        """Return the bounds fore each parameter a 2 x N array"""
-        limits = [p.bounds.limits for p in self._parameters]
+        """Return the bounds for each parameter as a 2 x N array"""
+        limits = [p.prior.limits for p in self._parameters]
```
pkienzle, line 339: bounds can be tricky, coming from the product of all priors, including both the hard limits from any parameter linked to the free variable as well as any limits set by the user.

bumps/fitproblem.py
```diff
@@ -515,156 +478,87 @@ def __call__(self, pvec=None):
...
+        for p in all_parameters:
+            # slot = p.slot
+            # value = p.value
+            
+            # TODO: this is a shim to accomodate Expression, Calculation etc. being
+            # put into attributes that have type "Parameter" (in user scripts)
+            # Do we cause those scripts to break instead?  
+            # Or do we autoconvert to .equals()?
+            if hasattr(p, 'add_prior'):
+                p.add_prior(p.distribution, bounds=p.bounds, limits=p.limits)
```
pkienzle, line 520: The interface is suspicious: `add_prior` is being called with a bunch of information it already has (`p.distribution`, `p.bounds`, `p.limits`). Also, modifying the state of an active object (the parameter) is best avoided.

There may be a simpler algorithm for coalescing the limits across all parameters.

bumps/fitproblem.py
```diff
@@ -515,156 +478,87 @@ def __call__(self, pvec=None):
...
+             broken.append("user constraint function is unsatisfied")
+        warnings.warn("Unsatisfied constraints: %s" % (",\n".join(broken)))
+        self.broken_constraints = broken
+
+        # TODO: shimmed to allow non-Parameter in Parameter attribute spots.
+        pars = []
+        for p in all_parameters:
+            slot = getattr(p, 'slot', None)
+            if isinstance(slot, Variable) and not p.fixed:
+                pars.append(p)
```
pkienzle, line 545: Delegate to parameter whether or not it contains a variable; move this code to parameter.py.

bumps/fitproblem.py
```diff
@@ -515,156 +478,87 @@ def __call__(self, pvec=None):
...
+        self._parameters = pars
+        # TODO: shimmed to allow non-Parameter in Parameter attribute spots.
+        self._bounded = [p for p in all_parameters if hasattr(p, 'has_prior') and p.has_prior()]
```
pkienzle, line 548: Check how this is used: maybe we just need the bounds for the fitted parameters to give to the optimizer?

bumps/fitproblem.py
```diff
@@ -676,16 +570,27 @@ def model_update(self):
...
         # subspace.  SrFit does this.
         for f in self.models:
-            f.model_update()
+            if hasattr(f, 'update'):
+                f.update()
 
     def model_nllf(self):
         """Return cost function for all data sets"""
-        return sum(f.model_nllf() for f in self.models)
+        return sum(f.nllf() for f in self.models)
```
pkienzle, line 578: model_nllf, constraints_nllf, etc. are internal methods that should be prefixed by underscore.

bumps/fitproblem.py
```diff
@@ -143,22 +149,34 @@ def plot(self, view='linear'):
...
-def no_constraints():
+def no_constraints() -> float:
     """default constraints function for FitProblem"""
     return 0
 
-
-# TODO: refactor FitProblem definition
-# deprecate the direct use of MultiFitProblem
-def FitProblem(*args, **kw):
+def chisq_str(fitness: Fitness) -> str:
```
pkienzle, line 156: It is confusing to have chisq_str() as a model function and a separate function chisq_str() as a method in the FitProblem class.

pkienzle:
`dataclasses` requires pip-installed backport for python 3.6. Are we dropping 3.6 support and moving to 3.7 (numpy moved in January)? Or do we hack setup.py to include dataclasses in the dependencies when python version is less than 3.7?

bumps/bounds.py
```diff
@@ -186,7 +202,7 @@ def init_bounds(v):
         return Bounded(lo, hi)
 
 
-class Bounds(object):
+class Bounds:
```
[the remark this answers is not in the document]
pkienzle: Okay, I see what you've done now: you added additional classes Uniform, Normal, SoftBounded in parameter.py and transform these to truncated variants using add_prior(). Since add_prior() is being called on reset, they don't need to be serialized, and so don't need the `@schema` decorator.

It may be less confusing to pull the frozen and unfrozen versions of the distribution to the same file.

pkienzle:
The existing `add_prior` doesn't support the discovery of the minimal bounding box for the random variables in a fit. While not necessary for correct fitting results, a tighter bounding box makes the fit more efficient, so it is worth a bit of effort to shrink it.

Direct bounds can come from natural limits on the prior, from truncated limits provided by the user, or from hard limits imposed by the model.

Priors have a natural, possibly definite domain. For example, lognormal operates on [0, ∞) and beta on [0, 1], but normal is unbounded. Practically, we can bound an indefinite range using the inverse cdf over [-ε, 1-ε] for small ε (e.g, 1e-3 or 1e-5). That is, we would require extraordinary evidence to a parameter 5σ away from the prior value, and the chances are very good that we will adjust the model if that is happening instead of accepting it. Bumps will be much happier with a finite range.

Users can additionally supply bounds [a, b] to form a truncated prior.

Hard limits are set by the model, such as thickness > 0.  [An aside: models could set a prior distribution on the parameter when they define them. I'm not sure what the consequences are for doing this. As it stands, the prior will be replaced by uniform when the user sets the fitted range.]

Combined, the direct bounds on a parameter are (p ∈ prior limits ∩ user limits ∩ parameter limits). This is easy to calculate and provides a simple upper bound on the bounding box.

The next challenge is to take parameter equality into account. If two parameters are equal then they both have to satisfy both limits. For example

```python
    a = Parameter(limits=[0, ∞))
    b = Parameter(limits=(-∞, ∞))
    a.dev(μ=3, σ=2, limits=[-6, 6])
    b.equals(a)
    b.range(1, 10)
```

Bounds on `a` are thus:

    a ∈ [0, ∞) ∩ [3-5⋅2, 3+5⋅2] ∩ [-6, 6]  ∩ (-∞, ∞) ∩ [1, 10] = [1, 6]

The bounds on `b` don't matter since it is not a fitted parameter, and we've already restricted `a` so that `b` will be satisfied. Since `b` is uniform and in range the contribution to the nllf is a constant and can be ignored.  We do have to evaluate the nllf for the non-uniform prior `a` and add it to the total nllf. This is true even if we reverse the equality so that `a.equals(b)`.

The above extends to a chain of equalities, each with their own set of bounds. The limits on every member of the chain is the intersection of each member of the chain. However, you only care about the limits on the random variables that are being fitted, so in practice you only need to propagate down the chain, not up.

It is hard to propagate constraints through expressions. If the total thickness is constrained how does that propagate down to constraints on the individual thickness? Given `a>0` and `b>0` and `a+b = 10 ± 1`, then `a+b ∈ [5, 15]` and so `a, b ∈ [0, 15]`. Even if we do create this restriction we still have to test the condition every time we evaluate the expression so it wouldn't save us any computation even if the prior on `a+b` was uniform. I think we punt on propagation; users can restrict the individual parameters `a, b` themselves if they want a more efficient fit. A simple subset is solvable (`constant*a + constant`), but I don't think it occurs enough to be worth supporting given that we are already supporting the more general case.

Note that long equality chains are expensive to evaluate. We need to dereference the slot then dereference the slot etc all the way to the base random variable. If we do this with a simple recursive algorithm then the cost for the model to get the value for each parameter is `k²` for a chain of length `k`.

So the preprocessing before the fit:

1. check for cycles in the constraints
2. evaluate all constraints, keeping a list of those not satisfied
3. display any failed constraints and exit if the starting point is outside the feasible region
4. gather all random variables which are not fixed as `fitted`
5. set the direct limits on the random variables 
6. propagate limits through equality chains to the base random variable or expression
7. gather all non-linear priors and all expressions with constraints as `constraints`

During the fit:

1. set the fitted parameter values
2. sum all `constraints`
3. if feasible, evaluate model
4. return constraints nllf plus model nllf

Propagating limits is a bit tricky if we start working about chain lengths and k² algorithms. Since this only happens at the start of the fit, lets assume that the chain length is short enough that we don't need to worry about it. Then we can use something like the following for propagating parameters and gathering constraints.

```python
    model_parameters = self.parameters()  # after renaming FitProblem.model_parameters to FitProblem.parameters
    parameters = unique(model_parameters)
    bounds = {p: (-∞, ∞) for p in parameters}
    for p in parameters:
        # apply parameter limits to whatever bounds are known for the parameter
        limits = p.prior.limits ∩ p.limits ∩ p.bounds
        slot = p.slot
        while isinstance(slot, Parameter):
            slot = p.slot
        bounds[p] = bounds[b] ∩ limits
    fitted = [p for p in parameters if p is not fixed]
    bounding_box = [bounds[p] for p in fitted]
    constraints = [p for p in parameters if p.prior is not uniform or isinstance(p.slot, Expression) and bounds[p] != (-∞, ∞)]
```

Yet another complication: how does all this interact with `FreeVariables`?

For example:
```python
    b.thickness.equals(a.thickness)
    b.dev(10, 1)
    free = FreeVariables([model names], a=a.thickness)
    free.a[0].range(1, 10)
    free.a[1].range(9, 20)
```
For bounds should have `free.a[0] ∈ [5,10], free.a[1] ∈ [9, 15]`. We might be able to make this happen by having an implicit `free.a[k].equals(a.thickness)` while propagating equality constraints. This could be in addition to an actual `free.a[k].equals(p)` for some other parameter `p`.

Even worse,
```python
    b.thickness.equals(a.thickness)
    b.dev(10, 1)
    free = FreeVariables([model names], a=a.thickness, c=c.thickness)
    free.a[0].range(1, 10)
    free.a[1].range(9, 20)
    free.c[0].equals(b.thickness)
```

Each constraint (i.e., the prior nllf) needs to be reevaluated separately for each model. This is obvious in the above example where `b.thickness.dev` will be evaluated once for each `free.a`. Even if `b.thickness` were a random variable independent of `a.thickness` I think it still makes sense to reevaluate it for each set of free parameters so you get some continuity between simultaneous fitting and batch fitting.

During the fit with free variables:

1. set the fitted parameter values
2. for each dataset:
3.     substitute free variables
4.     sum all `constraints`
5.     if feasible, evaluate model
6.     accumulate constraints nllf and model nllf
7. return constraints nllf plus model nllf 

We should consider caching evaluated parameters to reduce the cost of long chains and big expressions. Borrowing code from sasview, we can construct the dependency graph for the parameters. Put a `_value` attribute on each parameter then build a function which sets `p._value` in the order of dependency.

The `Calculation` class will need a parameters() method to return the dependencies. This will need to be a callback to support the legacy direct assignment of parameters rather than `.equals()`. That is, the dependent parameters may change between calls to `model_reset()`.

A nasty complication: `FreeVariables`. We need to reevalute derived parameters within each model whenever we substitute the set of free parameters. This is true for constraints such as `p.prior.nllf()` as well. In light of this I suggest holding off on caching calculations until we've done some profiling to see if it is necessary.

bumps/bounds.py
```diff
@@ -66,8 +66,24 @@
     # failure if it doesn't exist.
     pass
 
+from .util import field, schema, Optional, Any, Union, Dict, Callable, Literal, Tuple, List, Literal
```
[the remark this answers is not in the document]
bmaranville: Because of support for Python 3.6 - otherwise many modules would need to do the dance of trying to import e.g. Literal, failing, then importing it from typing_extensions

bumps/fitproblem.py
bmaranville: This is the collision of property with dataclasses... it is easily solved, just have to make a separate schema class.  I tried out both patterns, and if this one will cause issues we can definitely just use the inherited schema (I think that's used almost everywhere else)

bumps/fitproblem.py
bmaranville: The list that is generated here is definitely used as if it contained only fitted parameters (those with Variables).  Including all the parameters would cause breakage when it gets used, unless you want to filter the list every time?

bumps/fitproblem.py
bmaranville: I don't understand - before the refactor this list (_bounded) contained only objects that had a Bounds object defined (exculding Unbounded.)  It is doing the same now, where has_prior provides the same information.  We could rename it, but after looking through all the code what is expected from this iist is that it is every parameter that has a prior probability defined.  Rename to self._with_prior?

bumps/fitproblem.py
bmaranville: True, when they both take a Fitness instance as their first argument (self in one case)

bumps/fitproblem.py
bmaranville: > `dataclasses` requires pip-installed backport for python 3.6. Are we dropping 3.6 support and moving to 3.7 (numpy moved in January)? Or do we hack setup.py to include dataclasses in the dependencies when python version is less than 3.7?

It's not much of a hack, since there is a standardized way to do this.

bmaranville:
I would agree on deferring the caching implementation until we see if it is needed... I don't think users are clamoring for the ability to create long chains of constraints.  Most constraints (99%?) will continue to be equality constraints with Uniform priors, just like they make now with "assignment constraints".  There is some enthusiasm for easy-to-use inequalities, but these will be pretty short chains too.  The requirement that the user be able to understand (and explain in words!) every constraint used puts some... constraints on constraints usage.

bumps/fitproblem.py
bmaranville: It occurred to me when writing this that the logic could just become a method on Parameter, but that could be done anytime, and doesn't really change anything.

bmaranville:
People are asking about easier constraints... any chance we can start working on a merge for this soon?

Issue #65, opened 2021-05-27
Curve cannot handle decorated functions


hoogerheide:
Using a decorated function (e.g. jit) as the input to Curve generates an "unsupported callable" error. This occurs because Curve.__init__ calls inspect.getfullargspec which cannot handle decorated functions either. A potential workaround for inspect is in this thread: https://github.com/lmfit/lmfit-py/issues/570#issuecomment-516637403, but the overall recommendation appears to be to move to inspect.signature instead.

Note that calling the jit-ted function with another function does not give this error (e.g. def fitfunc(x, m, b): return line(x, m, b)), but does produce a 25-50% performance hit, at least on my machine. Using lambda in the Curve definition also avoids this error, but I haven't characterized performance yet.

bmaranville:
Can you post your affected model?

hoogerheide:
Meant to attach with the original post. Here it is.
[curve_jit.py.txt](https://github.com/bumps/bumps/files/6564978/curve_jit.py.txt)

bmaranville:
This replaces arginspect with signature: but now there is a new error, that says numba cannot compile line.  I throw it back over the fence to you, @hoogerheide 
```diff
diff --git a/bumps/curve.py b/bumps/curve.py
index 7457e77..72433cf 100644
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -74,7 +74,14 @@ def _parse_pars(fn, init=None, skip=0, name=""):
     *pk=None* then the parameter will be non-fittable, and instead set
     through *init*.
     """
-    pnames, vararg, varkw, pvalues = inspect.getargspec(fn)
+    sig = inspect.signature(fn)
+    params = sig.parameters.values()
+    #pnames, vararg, varkw, pvalues = inspect.getfullargspec(fn)
+    pnames = [p.name for p in params]
+    pvalues = [p.default for p in params]
+    vararg = [p for p in params if p.kind.name == 'VAR_POSITIONAL']
+    varkw = [p for p in params if p.kind.name == 'VAR_KEYWORD']
+
     if vararg or varkw:
         raise TypeError(
             "Function %r cannot have *args or **kwargs in declaration"
@@ -88,11 +95,9 @@ def _parse_pars(fn, init=None, skip=0, name=""):
     defaults = dict((p, 0) for p in pnames)
 
     # If the function provides default values, use those.
-    if pvalues:
-        # Ignore default value for "x" parameter.
-        if len(pvalues) > len(pnames):
[...]

PR #66, opened 2021-06-01
Allow numba models

bmaranville:
Apply some updates, replacing deprecated inspect.getargspec with inspect.signature and apply a __name__ attribute to the DynamicPackage created when the model is loaded.  These two fixes together enable @numba.jit functions in a user model.

Addresses #65

### DIFF
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -74,7 +74,14 @@ def _parse_pars(fn, init=None, skip=0, name=""):
     *pk=None* then the parameter will be non-fittable, and instead set
     through *init*.
     """
-    pnames, vararg, varkw, pvalues = inspect.getargspec(fn)
+    sig = inspect.signature(fn)
+    params = sig.parameters.values()
+    #pnames, vararg, varkw, pvalues = inspect.getfullargspec(fn)
+    pnames = [p.name for p in params]
+    pvalues = [p.default for p in params]
+    vararg = [p for p in params if p.kind.name == 'VAR_POSITIONAL']
+    varkw = [p for p in params if p.kind.name == 'VAR_KEYWORD']
+
     if vararg or varkw:
         raise TypeError(
             "Function %r cannot have *args or **kwargs in declaration"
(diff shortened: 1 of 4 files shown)


bmaranville:
Apparently there is no inspect.signature in python 2.7.  Can we finally drop support for 2.7?

bumps/curve.py
```diff
@@ -74,7 +74,14 @@ def _parse_pars(fn, init=None, skip=0, name=""):
...
     """
-    pnames, vararg, varkw, pvalues = inspect.getargspec(fn)
+    sig = inspect.signature(fn)
+    params = sig.parameters.values()
+    #pnames, vararg, varkw, pvalues = inspect.getfullargspec(fn)
+    pnames = [p.name for p in params]
+    pvalues = [p.default for p in params]
+    vararg = [p for p in params if p.kind.name == 'VAR_POSITIONAL']
+    varkw = [p for p in params if p.kind.name == 'VAR_KEYWORD']
+
```
pkienzle, line 84: Maybe use the signature object directly in the code below. It requires a little more thought than reconstructing `getargspec` but should be cleaner overall.

For example:
```python
valid = [p.kind in (Parameter.POSITIONAL_ONLY, Parameter.POSITIONAL_OR_KEYWORD) for p in params]
if not all(valid):
    raise TypeError(f"Only positional and keyword arguments allowed for {fn.__name__}")
```

pkienzle:
Yes we can drop 2.7 support. I'm not sure about 3.6—just how conservative is RHEL?

bmaranville:
RHEL8 seems to have python3.6, and it looks like our local RHEL7 computers have been upgraded to that as well.

pkienzle:
I don't remember when f-strings were introduced. We probably can't use them yet. We do need the equivalent of the above logic to protect against `def f(..., /, ...)` syntax for defining non-keyword positional arguments. Or maybe we allow them, but we do need to watch out for the `/` in the function signature.

bmaranville:
Ok, I'll make the change as above.  F-strings are from python3.6 forward so we can use them now!

pkienzle (approved):
I haven't tested, but the code looks okay. I would be tempted to fix my poor naming (clearly I should have used `module_name` rather than `package`, and maybe `DynamicModule` rather `DynamicPackage`) but not necessary.

pkienzle:
As you wish. I've decided that software projects are like gardens: always new things to plant and weeds to pull. No need to get everything in one pass.

bmaranville:
If creeping charlie finds some way to invade our code base, I shall abandon it entirely.

PR #68, opened 2021-06-15
add store, name options to fit call

cyankaet:
Allow user to specify a store location for dream summary graphics from the fit() interface in-script

### DIFF
--- a/bumps/fitters.py
+++ b/bumps/fitters.py
@@ -1222,6 +1222,34 @@ FIT_DEFAULT_ID = SimplexFit.id
 assert FIT_DEFAULT_ID in FIT_ACTIVE_IDS
 assert all(f in FIT_AVAILABLE_IDS for f in FIT_ACTIVE_IDS)
 
+def save_best(fitdriver, problem, best, store, name, view=None):
+    """
+    Save the fit data, including parameter values, uncertainties and plots.
+
+    *fitdriver* is the fitter that was used to drive the fit.
+
+    *problem* is a FitProblem instance.
+
+    *best* is the parameter set to save.
+    """
+    # Make sure the problem contains the best value
+    # TODO: avoid recalculating if problem is already at best.
+    problem.setp(best)
+    # print "remembering best"
+    problem.store = store
+    problem.name = name
+    problem.output_path = problem.store + '/' + problem.name
+    pardata = "".join("%s %.15g\n" % (name, value)
+                      for name, value in zip(problem.labels(), problem.getp()))
+    open(problem.output_path + ".par", 'wt').write(pardata)
+
+    fitdriver.save(problem.output_path)
+    with util.redirect_console(problem.output_path + ".err"):
+        fitdriver.show()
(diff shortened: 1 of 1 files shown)


cyankaet:
We moved the save_best() function because it was easier to manage the relevant imports, but almost certainly there is a way to refactor this change into the existing structure. We ran into circular import issues when we attempted.

pkienzle:
As you noted, the duplicated `save_best()` function is not ideal. The `fit()` function is already a hack, repeating bits of the command line interface. It would be best to reshuffle the code in `cli.main()`, creating a fit helper than can be called from either the command line or from the function interface, but that's too much for this PR.

You can avoid the additional duplication by moving `save_fit()` as-is into `fitters.py` and importing it into `cli.py`. You can populate the required fields (`problem.name`, `problem.store`, `problem.output_path`) from `fit()` before calling it. It would be better to move `fit()` into CLI but that changes existing public interfaces, so best not do that here.

You can hide this logic in a `fit_save()` function. Something like:
```python
def _fit_save(fitdriver,  x, **options):
    """
    Save the fit results and plots to a directory.

    Note: updates fitdriver.problem with store, name and output_path.
    """
    problem = fitdriver.problem

    # Save the plot if a store directory is given, either using the store='path'
    # argument or using problem.store if the store argument is not given.
    if 'store' in options:
        problem.store = options['store']
    store = getattr(problem, 'store', None)
    if not store:
        return

    # Similar to "store", the name of the problem can be passed as an argument
    # or given in the problem. It defaults to 'model'.
    if 'name' in options:
        problem.name = options['name']
    name = getattr(problem, 'name', 'model')

    # Allow the caller to select 'linear', 'log', or whatever view the plotter accepts.
    view = options.get('view', None)

    # Make the save directory. Unlike the command line we are not warning if it already exists.
    # TODO: duplicated from cli.make_store().
    problem.output_path = os.path.join(store, name)
    if not os.path.exists(store):
        os.mkdir(store)

    # Save the output.
    save_best(fitdriver, problem, x, view=view)
```

The `options` dict is being passed to the various fitters; it is not intended to have non-fit options inside. Although chances are good that unknown options will be ignored, you can guarantee this by modifying `_fill_defaults` to only include values that are represented in `settings`:
```python
result = dict(settings)
result.update((k, v) for k, v in options.items() if k in settings)
```

You may want to suppress the terminal output from the call to `fitdriver.show()` in `save_best()`, perhaps using the `verbose` flag.

bmaranville:
@acaruana2009 how does this fit with your recent work on model saving?  I can't locate the branch where you're working.

acaruana2009:
The branch I am working on is here: [https://github.com/bumps/bumps/tree/model_uncertainty_update_ajc/bumps](url)

I have just had a look and I have not fully completed my save out module yet.
At first glance Pauls suggestion should still work in my reworked branch but I would have to test it. If it is useful feel free to merge the above and I will try to rebase my branch and test to make sure it still works (just need to remember the best way of getting my branch up to date with master).

PR #76, opened 2021-08-23
add comments regarding the different kinds of uncertainty analysis

pkienzle:
Soap-boxing about the preference for credible intervals rather than MLE when performing uncertainty analysis.

Please read it in context and adjust the text in a way that users of the software will best understand.

### DIFF
--- a/doc/guide/fitting.rst
+++ b/doc/guide/fitting.rst
@@ -84,13 +84,61 @@ Uncertainty Analysis
 ====================
 
 More important than the optimal value of the parameters is an estimate
-of the uncertainty in those values.  By casting our problem as the
-likelihood of seeing the data given the model, we not only give
-ourselves the ability to incorporate prior information into the fit
-systematically, but we also give ourselves a strong foundation for
+of the uncertainty in those values. The particular optimum is an accident
+of the measurement; perform the measurement again and you will get a
+different optimum. Given the uncertainty in the measurement, there is a
+joint distribution of parameter values that are consistent with the
+measurement. The goal of uncertainty analysis is to determine this
+distribution and summarize it for the reader.
+
+By casting our problem as the likelihood of seeing the data given the model,
+we not only give ourselves the ability to incorporate prior information into
+the fit systematically, but we also give ourselves a strong foundation for
 assessing the uncertainty of the parameters.
 
-Uncertainty analysis is performed using :ref:`fit-dream`.  This is a
+There are multiple ways to perform the analysis:
+
+1. Bayesian inference. Given the probability on the parameters and the
+   probability that the measured data will be seen with those parameters,
+   infer the probability of the parameters given the measured data.  This
+   is the primary method in Bumps and will be discussed at length below.
+2. Sensitivity analysis. Given the optimal parameter values, look at the
+   curvature around that point as a Gaussian distribution with covariance
+   computed from the Hessian matrix. Further, pretend that there is no
+   interaction between the parameters, and report the uncertainty as the
+   square root of the diagonal. This is the default method for most optimizers
+   in Bumps.
+3. Uncertainty contour. Assuming the measurement data is independent and
+   normally distributied, a given increase in $\chi^2$ above the minimum
+   corresponds to 1-$\sigma$ confidence interval. By following this contour
+   you can find the set of all points $\xi$ such that
+   $\chi^2(\xi) = \chi^2(x) + C$ where $x$ is the point of maximum
+   likelihood. Look in Numerical Recipes chapter on nonlinear least squares
+   for a more complete discussion.  Bumps does not include algorithms
+   for this kind of analysis.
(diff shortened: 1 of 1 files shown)


doc/guide/fitting.rst
```diff
@@ -84,13 +84,61 @@ Uncertainty Analysis
...
 
 More important than the optimal value of the parameters is an estimate
-of the uncertainty in those values.  By casting our problem as the
-likelihood of seeing the data given the model, we not only give
-ourselves the ability to incorporate prior information into the fit
-systematically, but we also give ourselves a strong foundation for
+of the uncertainty in those values. The particular optimum is an accident
+of the measurement; perform the measurement again and you will get a
+different optimum. Given the uncertainty in the measurement, there is a
+joint distribution of parameter values that are consistent with the
```
acaruana2009, line 90: Could we expand on the join distribution slightly? 

Is the joint distribution, the combined distributions from both the simulations with respect to the data and the error in the data? Or have I missed what this is completely?

doc/guide/fitting.rst
```diff
@@ -84,13 +84,61 @@ Uncertainty Analysis
...
+the fit systematically, but we also give ourselves a strong foundation for
 assessing the uncertainty of the parameters.
 
-Uncertainty analysis is performed using :ref:`fit-dream`.  This is a
+There are multiple ways to perform the analysis:
+
+1. Bayesian inference. Given the probability on the parameters and the
+   probability that the measured data will be seen with those parameters,
+   infer the probability of the parameters given the measured data.  This
+   is the primary method in Bumps and will be discussed at length below.
```
acaruana2009, line 104: Is it worth explicitly mentioning that this is currently only done with DREAM, i.e. from MCMC sampling (or again do we do this with the other fitters in some way)?

doc/guide/fitting.rst
```diff
@@ -84,13 +84,61 @@ Uncertainty Analysis
...
+2. Sensitivity analysis. Given the optimal parameter values, look at the
```
acaruana2009: The use of "optimal" here might be confusing given the use of "optimum" in the introducing paragraph. See my edits for clarity to see how I have interpreted this.

doc/guide/fitting.rst
```diff
@@ -84,13 +84,61 @@ Uncertainty Analysis
...
+   curvature around that point as a Gaussian distribution with covariance
+   computed from the Hessian matrix. Further, pretend that there is no
+   interaction between the parameters, and report the uncertainty as the
+   square root of the diagonal. This is the default method for most optimizers
+   in Bumps.
+3. Uncertainty contour. Assuming the measurement data is independent and
+   normally distributied, a given increase in $\chi^2$ above the minimum
```
acaruana2009, line 112: For consistency, we may want to either use "normal" or "gaussian" for describing the distribution here, rather than switching between the two. Since in the previous point Gaussian is used, a non-expert reader might become slightly confused that we are talking about different things - I am aware that it *should* be obvious that these are the same thing (again unless I have missed a subtle designation between the nomenclature).

doc/guide/fitting.rst
```diff
@@ -84,13 +84,61 @@ Uncertainty Analysis
...
+   corresponds to 1-$\sigma$ confidence interval. By following this contour
+   you can find the set of all points $\xi$ such that
+   $\chi^2(\xi) = \chi^2(x) + C$ where $x$ is the point of maximum
+   likelihood. Look in Numerical Recipes chapter on nonlinear least squares
```
acaruana2009, line 116: Is there a direct reference we can point the readers to here?

doc/guide/fitting.rst
```diff
@@ -84,13 +84,61 @@ Uncertainty Analysis
...
+   for a more complete discussion.  Bumps does not include algorithms
+   for this kind of analysis.
+4. Forward Monte Carlo. Bumps has the option :ref:`option-resynth` to perform
+   a forward Monte Carlo estimate of the maximum likelihood.  That is, you
+   can use the measurement uncertainty to "rerun" the experiment, synthesizing
+   a new dataset with the same uncertainty but slightly different values
+   and finding the maximum likelihood. The result after $n$ runs is a sample
+   similar to Bayesian inference but it is from a joint distribution of
+   maximum likelihood rather than a joint distribution of likelihood. This
```
acaruana2009, line 125: I am wondering if it would be useful to have a section that briefly describes the difference between the "maximum likelihood" and the "likelihood" in the context of the joint distribution. 

In general, the likelihood function (especially in the form of the nllf that is used for optimisation), is an important concept that new users to optimisers are often not familiar with. There is of course a lot of literature for users to read about this, but is it worth us giving a very brief introduction somewhere and then point them to a useful reference?

acaruana2009 (commented):
See my comments and my edits. I hope this helps. 

I have just added points where I feel either that a user would not easily pick up what we are trying to say, or where I am not sure what is being said and could benefit from a little further explanation.

PR #79, opened 2021-11-03
apply string formatting (%.10g) to values in summary view

bmaranville:
Floating-point values in summary view were overflowing the columns for "value".  This applies a uniform %.10g string format to all the values in the view.

### DIFF
--- a/bumps/gui/summary_view.py
+++ b/bumps/gui/summary_view.py
@@ -153,7 +153,8 @@ class SummaryView(scrolled.ScrolledPanel):
         for p in self.display_list:
             p.update_slider()
 
-VALUE_FORMAT = "{:.10g}"
+VALUE_PRECISION = 6
+VALUE_FORMAT = "{{:.{:d}g}}".format(VALUE_PRECISION)
 
 class ParameterSummary(wx.Panel):
     """Build one parameter line for display."""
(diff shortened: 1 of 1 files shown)


pkienzle:
Yeah, I saw the ugliness. Given that the slider only has 3 or 4 digits of precision at best ten digits feels like overkill. How about `.5g`? Trying it with a range of values in py3.8 it seems to do the right thing.

If we want to be pedantic, we should set the parameter value to the displayed value converted to float so we get consistent results from extremely sensitive parameters. Maybe only after touching the slider so that the initial curve uses the fitted parameters when a .par file is loaded.

bmaranville:
Agreed that adjusting the parameter value should only happen if the slider is touched/moved, but as far as I can tell this is already the behavior that occurs.  I used .10g because there's space for it, so why not?

pkienzle:
Particularly, adjust the parameter value to the formatted value shown after the parameter is adjusted. That means if you are clipping the last few digits for display you should be setting the parameter value without the last few digits. I do not know that it is or is not doing this already.

Re: .10 vs .5, whatever is aesthetically best. I would prefer to only report as many digits as we have control over from the slider, but if you prefer .10 I do not object (he who writes the code wins the toss).

pkienzle (approved):
Code looks fine. I didn't test.

bmaranville:
I would like to implement your suggestion of setting parameter value to "nice" formatted value on slider move, @pkienzle, but I'm having a weird problem: I don't see where in summary_view.py the value is actually being changed by the slider, except when OnScroll is triggered.  What am I missing?

bmaranville:
I am going to hold off on setting the parameter value to

bmaranville:
...pushed the wrong button...

PR #81, opened 2021-11-10
add a 'source' parameter to signal.update_model

bmaranville:
...so that updates can be ignored for same-source

An alternative to this is to add an argument to every view's update_model method that contains extra information, like the source of the event, and the update_model methods could then use that information if they want.

In this version, if you pass a "source" parameter to signal.update_model,  bumps.app_panel will not run view.update_model for any view where the title matches. "source"

### DIFF
--- a/bumps/gui/app_panel.py
+++ b/bumps/gui/app_panel.py
@@ -337,9 +337,9 @@ class AppPanel(wx.Panel):
     def OnModelNew(self, model):
         self.set_model(model)
 
-    def OnModelChange(self, model):
+    def OnModelChange(self, model, source=None):
         for v in self.view.values():
-            if hasattr(v, 'update_model'):
+            if getattr(v, 'title', '') != source and hasattr(v, 'update_model'):
                 v.update_model(model)
 
     def OnModelSetpar(self, model):
--- a/bumps/gui/parameter_view.py
+++ b/bumps/gui/parameter_view.py
@@ -36,6 +36,7 @@ from .util import nice
 from . import signal
 
 IS_MAC = (wx.Platform == '__WXMAC__')
+VIEW_NAME = "Parameters"
 
 class ParameterCategory(object):
     def __init__(self, name):
@@ -219,13 +220,13 @@ class ParametersModel(dv.PyDataViewModel):
             # update; parameter values didn't change, so dirty=False, and
             # model.model_update() will not be called and the theory value
             # will not be re-computed.
-            signal.update_model(model=self.model, dirty=False)
+            signal.update_model(model=self.model, dirty=False, source=VIEW_NAME)
         else:
             signal.update_parameters(model=self.model, delay=1)
         return True
 
 class ParameterView(wx.Panel):
-    title = 'Parameters'
+    title = VIEW_NAME
     default_size = (640,500)
     def __init__(self, *args, **kw):
         wx.Panel.__init__(self, *args, **kw)
--- a/bumps/gui/signal.py
+++ b/bumps/gui/signal.py
@@ -20,15 +20,16 @@ def model_new(model):
     wx.CallAfter(send, 'model.new', model=model)
 
 
-def update_model(model, dirty=True):
+def update_model(model, dirty=True, source=None):
     """
     Inform all views that the model structure has changed.  This calls
     model.model_reset() to reset the fit parameters and constraints.
+    Include a source if debouncing is needed.
     """
     model.model_reset()  #
     if dirty:
         model.model_update()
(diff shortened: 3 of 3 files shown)


pkienzle:
I wonder if this fixes the problem: updates to the parameter window from whatever source scrolls the parameter window.

Perhaps the better solution is remember the scroll position before redrawing and restore it when done? For example, reloading a model probably triggers the redraw, though you don't necessarily want it to scroll.

I'm suspicious of adding a feature to a general interface that is only used in one place. Why don't other views and updates need debouncing? Or do they and we just don't know it yet?

Is the problem that we're pretending that toggling "Fixed?" is a structural change? I can see why we would do so: the summary view will need to include a new slider when that item is toggled. Maybe different solution is to treat a toggle as a normal model update, and let summary figure out if any fixed parameters have changed.

acaruana2009:
I don't know if this helps or confuses the issue more, but I have found on windows it does not jump around. 

I think with the summary view, it doesn't bounce if you play with the profile, but it does bounce if you add or remove a parameter in the parameter view. Of course it totally redraws if you reload a model.

bmaranville:
I looked into scrolling as a solution and it will not work.  Wx simply doesn't support it in this context.

bmaranville:
Also I suspect it will be easier to create interactive components if they are not forced to update themselves every time they broadcast an update to the rest of the app.  I would actually prefer to add the event source to every update_model but then we'd have to update every view and break the backwards compatibility of the API.

bmaranville:
I would like to look into the other suggestion though... It makes more sense to me to treat togglng the fit flag as a parameter update than a model change.

pkienzle:
A general approach is to maintain an internal state in each GUI component (the view) separate from the desired state for that component (the model). The view only needs to update when the internal state differs from the desired state. Then it shouldn't matter which window triggered the update. The model update versus model change event distinction was a lazy way of skipping the check that the structure changed when updating the view.

It should be easy enough to update summary so that it either redraws itself if any parameters are toggled, or adds/removes them from the existing table without a complete redraw. If not then feel free to merge your solution. It feels wrong to me but I can't point to any immediate harm.

bmaranville:
Yes in principle a MVC architecture would work but here we have view where we directly alter the internal state, and then the change event is emitted with the state, which then returns through the update_model method, triggering a redraw of the view with the same state that was sent by the view in the first place.  This seems unecessary.
Even that would be fine, except that reinitializing the items and then calling Cleared() on the the dvModel seems to lose scroll position on linux (to be honest I'm not sure why it doesn't on Windows or Mac)

pkienzle:
It makes it easier to reason about the code if you don't care where the signal originated. In particular, it catches instances where the trigger was not directly from the view.

I can't point to any examples where this happend in the current code which is why I'm not requesting changes to the PR. I can imagine instance of model change being triggered elsewhere (e.g., load previous fit state from the menu) but I can't immediately come up with a model update example.

Note that a running fit will trigger a model update, but there are so many problems associated with twiddling a model while fitting that I haven't tried to support it. I'm not sure what will happen if you click "fit?" on a parameter while a fit is running.

bmaranville:
It seems straightforward to alter SummaryView so that it triggers update_model if p.fixed is changed for a parameter in the model, but is the call to model.model_parameters() going to be too expensive to do on every parameter update?  We'd have to recreate the filtering being done in FitProblem.model_reset, at every parameter_update:

pars = parameter.varying(parameter.unique(model.model_parameters())

pkienzle:
The code I'm looking at is using model._parameters. If we store the current value in `SummaryView` then check if `self.current_parameters is model._parameters` in `update_parameters` then we would know that they have changed. We will have to trigger `model.model_reset()` any time "fixed?" is clicked; this is currently happening in `signal.update_model`.

This bakes in the notion that `FitProblem._parameters` is replaced during `model_reset()`, which is a little too intimate for my tastes. We aren't even supposed to access `_parameters` since it is private. Are there other views that care about the list of active parameters that need this logic?

Yet again I am reminded that GUI programming is a complete mess and there are no right answers.

pkienzle:
Summary of alternate approach:
* `SummaryView.set_model` needs `self.parameters = model._parameters`
* `SummaryView._update_parameters` needs
```
if self.parameters is not model._parameters:
    self._update_model()
else:
    …
```
* `ParametersModel.SetValue` needs
```
if col==0:
    self.model.model_reset()
signal.update_parameter(…)
```
Of course I haven't tested this and there may be subtleties that I've missed.

I leave it to the he who commits the code to decide which approach is best.

bmaranville:
I will implement your suggestion from the last comment.

bmaranville:
superseded by #86

PR #88, opened 2021-12-15
support saving of all parameters, not just fitted ones (also reloading)

bmaranville:
Related to request for refl1d (clearly filed on the wrong repository) https://github.com/reflectometry/refl1d/issues/145

### DIFF
--- a/bumps/cli.py
+++ b/bumps/cli.py
@@ -125,9 +125,13 @@ def save_best(fitdriver, problem, best, view=None):
     # TODO: avoid recalculating if problem is already at best.
     problem.setp(best)
     # print "remembering best"
-    pardata = "".join("%s %.15g\n" % (name, value)
-                      for name, value in zip(problem.labels(), problem.getp()))
-    open(problem.output_path + ".par", 'wt').write(pardata)
+    fitted_pardata = "".join("%s %.15g\n" % (p.name, p.value)
+                      for p in problem._parameters)
+    fixed_pardata = "".join("%s %.15g\n" % (p.name, p.value)
+                      for p in problem._fixed_parameters)
+
+    pstring = "%s\n# Unfitted (fixed) parameters:\n\n%s" % (fitted_pardata, fixed_pardata)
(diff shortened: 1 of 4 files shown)


bmaranville:
From @pkienzle (review was not working yesterday on github.com)
I don't know why github isn't allowing me to review.

Looking at the green line numbers (yours):

L128 cli:

For PR #88 (save all pars) the reason I'm using labels() and getp() is because bumps does not assume it is working with a FitProblem object with Parameter objects for the model, but instead is working with vectors of p and corresponding labels.

The reason is that I wanted to able to wrap methods similar to the scipy fit functions which work on vectors, and with "direct" log probability surfaces.

I'm not sure if the ability is still there. Some assumptions about the data structures in FitProblems may have snuck into FitDriver, and the lack of complete testing run continuously may not have caught it.

It is worth asking whether the direct interface is worth supporting given that it has been eroding over the years. I have no opinion at the moment.

The expedient solution is to provide FitProblem with methods for allp() and all_labels() and put off the decision until later.

Something has to change because the PR is currently using the "private" attributes  ._parameters and ._fixed_parameters.


L173 cli:

Comments are fine. They may already be implicitly supported since IIRC I ignore any labels that are not in the model, but better to be explicit.


L176 cli:

I suppose it is more robust, but I would be tempted the throw and error, or at least log a warning if a parameter line didn't parse. I suspect you do not need both L173 and L176 conditions.


L653 app_panel:

I don't know if you need to separate fixed and fitted parameters.

You might want to list the bounds on each fitted parameter in the .par file, and have the line parser ignore things after name value.

For derived parameters you might want to put the value and the equation instead of the bounds.


L1073 parameter:

If you are doing all parameters intermixed in the .par file then you only need unique, not fixed.

pkienzle:
Redo after reorg. Dump all parameters to the .par file and to the fit state in hdf.

In hdf the problem may be pickled and unrecoverable, so we can't rely on it to retrieve the parameters to the fit.

PR #96, opened 2022-06-22
Update Random123 to v1.14.0

pkienzle:
Checked on MacOS. Needs testing on Linux and perhaps windows.

I used:
```
$ python run.py doc/examples/peaks/model.py --fit=dream --store=/tmp/t1 --burn=1000 --samples=50000
```
I modified `bumps/dream/core.py:246` to verify that the dll was loaded and running:
```python
if dll:
    print("Using compiled stepper")
    ...
```


pkienzle:
Last I looked (which was awhile ago) submodules are not something that "just works". It looks like it still requires special commands to work with it.

I pulled the library into the repo because I didn't expect it to change too much for what I need. I'm not using the GPU abilities, and I don't use much of the API.

Rather than put it in a submodule I would prefer to have a small build script to fetch the library before doing the build command. We can pin the version with the commit hash. This is a little different from a submodule in that we are not trying to keep in sync with the submodule master, but instead updating to a newly released version on our own schedule.

bmaranville:
That sounds very reasonable.  My comment should have been less specific, e.g. "Should we do something other than copy the whole source tree into the bumps tree?"

pkienzle:
Shall we put that in a new ticket? Or do you want it done now.

I'm in the process of updating the installation docs for a new user which is why the topic has come up.

bmaranville:
Given that the build process for the extension is completely manual, I would think the other simple option is just to change the instructions from 
```bash
(cd bumps/dream && cc compiled.c -I ../../Random123/include/ -O2 -fopenmp -shared -lm -o _compiled.so -fPIC)
```
to 
```bash
(rm -rf random123 && git clone -b v1.14.0 https://github.com/DEShawResearch/random123.git && cd bumps/dream && cc compiled.c -I ../../random123/include/ -O2 -fopenmp -shared -lm -o _compiled.so -fPIC)
```

pkienzle:
TLDR; dump random123 and replace the c code with numba

A pygit2 dependency in the build system is inconvenient, though `build_requires` can probably handle it. It appears to have wheels for everything so hopefully it doesn't require anything special in the environment, such as a git installation.

`_compiled.so` is not actually a python extension, just a plain DLL which I access using FFI. We probably want to tell setup.py to use `py_limited_api`. It will probably end up in some unexpected place with a funky name, so we would need to provide code to find it.

I don't want to support a large fleet of per-architecture wheels. In theory we could include DLLs for the different architectures in one wheel and use arch name to load the right one at run time. It'll make the CI build more complicated.

Using pure python is fast enough unless you have a lot of cores. We could leave it as is, with only the people installing it on a cluster having to worry about the compile step.

We can replace `compiled.c` in numba, using the numpy RNG rather than Random123. The big issue is parallel random number generation, which generates a different random sequence each time even when seeded with the same number:
```python
@numba.njit(parallel=True)
def f(n=1000,seed=1):
     np.random.seed(seed)
     ret = np.empty(n)
     for k in numba.prange(n): ret[k] = np.random.rand()
     return ret
```
which produces different sequences when called with the same seed:
```
>>> f(10,1)
array([0.32416581, 0.01376633, 0.36520072, 0.00891795, 0.20365458,
       0.8977796 , 0.57098256, 0.93662088, 0.11728586, 0.34377   ])
>>> f(10,1)
array([0.82117327, 0.48014496, 0.953453  , 0.52278308, 0.66671965,
       0.53462661, 0.13237632, 0.66870434, 0.72748357, 0.09524524])
```

The difference is not just a timing/task switching effect as shown by sorting the output:
```
>>> np.sort(f(10,1))
array([0.06604697, 0.09315705, 0.15412378, 0.22964105, 0.34745319,
       0.54485166, 0.55639817, 0.58050955, 0.62621583, 0.72635018])
>>> np.sort(f(10,1))
array([0.1080033 , 0.25391153, 0.49734202, 0.54360628, 0.61251823,
       0.61992736, 0.70812219, 0.77455048, 0.88034806, 0.99155404])
```

Ideally we would be able to reproduce the same fit with the same seed but it is not essential.

pkienzle:
Agreed that dumping the entire new Random123 package into the repo makes for an ugly changelog. I'll withdraw this PR.

I created a ticket for numba conversion.

I'm slightly concerned about `git clone -b v1.14.0 …`. I don't believe branch tags are part of the blockchain. They may decide to remove the branch later leaving us with broken instructions. I suppose I could put the compiled c build into CI so we at least know that it happens.

Issue #80, opened 2021-11-05
constraint handling broken?


bmaranville:
When a constraint function is defined the simplex fitter seems to drive to values that violate the constraints, even when the penalty is 1e9, and the constraints are initially satisfied:

![image](https://user-images.githubusercontent.com/686570/140578066-8637a7d8-99cc-4ffa-81c2-328870d4a604.png)

The DE fitter reports strange errors when a constraint function is defined:
```python
/home/brian/.virtualenvs/refl1d/lib/python3.9/site-packages/numpy/core/fromnumeric.py:87: RuntimeWarning: overflow encountered in reduce
  return ufunc.reduce(obj, axis, dtype, out, **passkwargs)
/home/brian/dev/bumps/bumps/mystic/stop.py:129: RuntimeWarning: overflow encountered in square
  return math.sqrt(np.sum(abs(x)**2))
```

acaruana2009:
I can confirm that I have also seen this. I believe a while ago when I was asking about constraints I thought it wasn't working properly. I can confirm it doesn't work in DREAM either. I can try to dig out my last scripts where I tried this if it is useful.

bmaranville:
It is working as far as I can tell in the dataclass_overlay branch, in which we worked on constraints quite a bit.  Maybe worth deferring work on this until we merge that branch.

acaruana2009:
I think that is probably best. That way new users can get used to the new constraint system, once it is in place. Definitely no rush on this one.

PR #100, opened 2022-07-28
Mark inner loop variables as private for openmp in bounds calculations

llimeht:
In packaging bumps 0.9.0, we noticed that the dream/bounds.py tests were failing; the failures were not always identical but were always when the bounds were reflected or folded on the multidimensional test at `bounds.py:207`, with 1 or 2 of the vectors not having the bounds applied to them correctly, and so the `norm` not evaluating to `0`.

After much gnashing of teeth, we found that the failures often went away when running within the debugger,  could be suppressed entirely if the C bounds code was not used (by setting `dll=None` in the python module), and could also be suppressed by preventing OpenMP from doing any parallelisation (by setting `OMP_NUM_THREADS=1` in the environment.

Looking at the C functions `bounds_reflect`, `bounds_clip`, `bounds_fold`, they each contain nested loops that are being parallelised using `#pragma omp parallel for private(idx)`. However, for nested loops, the inner loop variable needs to be private, else the parallelised loops will interfere with each other and return early. OPENMP automatically marks any variables that are declared within the parallelised block as being private but if the inner loop variable is declared at the top of the function, then this does not happen. The attached patch just makes the inner loop variable thread-private too.

It appears that this failure has been around for a while; I can see in the packaging git log that the tests of the `bumps.dream` module were disabled back in 2018 with a commit message citing a failure at the same location; it was only in switching from nose to pytest for 0.9.0 that it came to light again.

I assume that OpenMP is automatically selecting only one thread on GitHub CI which is why the failure is not seen there.

Test logs below:

**Failing**
```
$ py.test -v --no-cov ./bumps/dream/bounds.py     
============================================== test session starts ==============================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0+repack -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: …/bumps, configfile: pytest.ini
plugins: cov-3.0.0
collected 1 item                                                                                                

bumps/dream/bounds.py::test FAILED                                                                        [100%]

=================================================== FAILURES ====================================================
_____________________________________________________ test ______________________________________________________

    def test():
        """bounds handlers test"""
        from numpy.linalg import norm
        from numpy import array
    
        bounds = list(zip([5, 10], [-inf, -10], [-5, inf], [-inf, inf]))
        v = np.ascontiguousarray([6, -12, 6, -12], 'd')
        for t in 'none', 'reflect', 'clip', 'fold', 'randomize':
            assert norm(make_bounds_handler(bounds, t).apply(v+0) - v) == 0
        v = np.ascontiguousarray([12, 12, -12, -12], 'd')
        for t in 'none', 'reflect', 'clip', 'fold':
            w = make_bounds_handler(bounds, t)
>           assert norm(w(array([v, v, v])) - w.apply(v+0)) == 0
E           assert 23.173260452512935 == 0
E            +  where 23.173260452512935 = <function norm at 0x7fd273866a70>((array([[ 10., -10.,  -5., -12.],\n       [ 10., -10.,  -5., -12.],\n       [ 12.,  12., -12., -12.]]) - array([ 10., -10.,  -5., -12.])))
E            +    where array([[ 10., -10.,  -5., -12.],\n       [ 10., -10.,  -5., -12.],\n       [ 12.,  12., -12., -12.]]) = <bumps.dream.bounds.ClipBounds object at 0x7fd272bde470>(array([[ 10., -10.,  -5., -12.],\n       [ 10., -10.,  -5., -12.],\n       [ 12.,  12., -12., -12.]]))
E            +      where array([[ 10., -10.,  -5., -12.],\n       [ 10., -10.,  -5., -12.],\n       [ 12.,  12., -12., -12.]]) = <built-in function array>([array([ 12.,  12., -12., -12.]), array([ 12.,  12., -12., -12.]), array([ 12.,  12., -12., -12.])])
E            +    and   array([ 10., -10.,  -5., -12.]) = <bound method ClipBounds.apply of <bumps.dream.bounds.ClipBounds object at 0x7fd272bde470>>((array([ 12.,  12., -12., -12.]) + 0))
E            +      where <bound method ClipBounds.apply of <bumps.dream.bounds.ClipBounds object at 0x7fd272bde470>> = <bumps.dream.bounds.ClipBounds object at 0x7fd272bde470>.apply

bumps/dream/bounds.py:207: AssertionError
============================================ short test summary info ============================================
FAILED bumps/dream/bounds.py::test - assert 23.173260452512935 == 0
=============================================== 1 failed in 3.00s ===============================================
```

**Failure suppressed by forcing single threading**
```
$ OMP_NUM_THREADS=1 py.test -v --no-cov ./bumps/dream/bounds.py
============================================== test session starts ==============================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0+repack -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: …/bumps, configfile: pytest.ini
plugins: cov-3.0.0
collected 1 item                                                                                                

bumps/dream/bounds.py::test PASSED                                                                        [100%]

=============================================== 1 passed in 2.83s ===============================================
```

**Test passing after adding this patch**
```
$ py.test -v --no-cov ./bumps/dream/bounds.py
============================================== test session starts ==============================================
platform linux -- Python 3.10.5, pytest-7.1.2, pluggy-1.0.0+repack -- /usr/bin/python3
cachedir: .pytest_cache
rootdir: …/bumps, configfile: pytest.ini
plugins: cov-3.0.0
collected 1 item                                                                                                

bumps/dream/bounds.py::test PASSED                                                                        [100%]

=============================================== 1 passed in 2.97s ===============================================
```

### DIFF
--- a/bumps/dream/compiled.c
+++ b/bumps/dream/compiled.c
@@ -398,7 +398,7 @@ void bounds_reflect(int Nchain, int Nvar, double pop[], double low[], double hig
     int k, p, idx;
 
     #ifdef _OPENMP
-    #pragma omp parallel for private(idx)
+    #pragma omp parallel for private(idx, k)
     #endif
     for (p=0; p < Nchain; p++) {
         for (k=0; k < Nvar; k++) {
@@ -421,7 +421,7 @@ void bounds_clip(int Nchain, int Nvar, double pop[], double low[], double high[]
     int k, p, idx;
 
     #ifdef _OPENMP
-    #pragma omp parallel for private(idx)
+    #pragma omp parallel for private(idx, k)
     #endif
     for (p=0; p < Nchain; p++) {
         for (k=0; k < Nvar; k++) {
@@ -441,7 +441,7 @@ void bounds_fold(int Nchain, int Nvar, double pop[], double low[], double high[]
     int k, p, idx;
 
     #ifdef _OPENMP
-    #pragma omp parallel for private(idx)
+    #pragma omp parallel for private(idx, k)
     #endif
     for (p=0; p < Nchain; p++) {
         for (k=0; k < Nvar; k++) {
@@ -472,7 +472,7 @@ void bounds_random(int Nchain, int Nvar, double pop[], double low[], double high
     int k, p, idx;
 
     #ifdef _OPENMP
-    #pragma omp parallel for private(idx)
+    #pragma omp parallel for private(idx, k)
     #endif
     for (p=0; p < Nchain; p++) {
         for (k=0; k < Nvar; k++) {


llimeht:
Possibly explains #80.

pkienzle:
#80 is referencing the simplex fitter which is using its own bounds handling code. The C code is isolated to DREAM.

Note that simplex as written isn't well suited to parallel since the first couple of checks move a single vertex, and only the shrink/expand step at the end computes all vertices. We did have promising results by checking multiple vertices in the move and reflect steps but that code never made it into production. Worth revisiting, but that is a separate issue.

pkienzle:
Code looks fine. Thanks!

Unfortunately my CI does not exercise the compiled module so I'll need to check by hand before pulling.

llimeht:
> #80 is referencing the simplex fitter which is using its own bounds handling code. The C code is isolated to DREAM.

There was also a mention of bounds failing for dream in #80 which is what I had picked up on there: "I can confirm it doesn't work in DREAM either". As you say, it can't be all of the story for #80.

PR #102, opened 2022-08-01
Run CI pytest against the compiled code too

llimeht:
As per comment in #100, the dream/compiled.c code is not currently tested on CI. This PR is to make that happen.

(Untested changes: see #101)


llimeht:
Test failure described in #100 is now exposed on CI, so you can use CI to check that #100 fixes the bug :smile:

PR #114, opened 2022-10-31
treat empty string args as no args 

bmaranville:
addresses issue with windows batch file args processing, where passing no arguments in sometimes results in an empty string being passed as a single argument.

### DIFF
--- a/bumps/cli.py
+++ b/bumps/cli.py
@@ -533,6 +533,10 @@ def main():
     """
     # add full traceback to warnings
     #warnings.showwarning = warn_with_traceback
+    if len(sys.argv) == 2 and sys.argv[1] == "":
+        # Windows batch files pass empty string as args if none specified;
+        # treat this as no arguments:
+        sys.argv.pop()
 
     if len(sys.argv) == 1:
         sys.argv.append("-?")


pkienzle:
Eww... I would like to know the range of the perversion before applying a fix.

For example, do we need to check all final arguments?
```python
if not sys.argv[-1]:
   sys.argv.pop()
```

Can it be anywhere on the line?
```python
sys.argv = [sys.argv[0], *(v for v in sys.argv[1:] if v)]
```

Both of these could happen if the argument processing is something like something like `"p1  p2 ".split(" ")` => `['p1', '', 'p2', '']` instead of `"p1  p2 ".split()` => `['p1', 'p2']`.

We don't currently have any options where a blank is valid, but I can imagine,  e.g., `--title ""` at the end of the command line. I guess we don't worry about that for now.

bmaranville:
Agreed - I should check for when this will be an issue.  My guess is that it is only when no arguments at all are specified.

bmaranville:
On further investigation this PR only fixes one cosmetic issue with the .bat file in question.  We should probably not make the fix here, and instead and the batch file needs to be refactored (refl1d.bat)

pkienzle:
BTW, the "eww" was regarding the bug, not the proposed fix 8-)

pkienzle:
So something like the following?
```batch
if "%~1" == "" (
    "%~dp0\python.exe" "%~dp0\Scripts\refl1d_cli.py"
) else (
    "%~dp0\python.exe" "%~dp0\Scripts\refl1d_cli.py" "%*"
)
```
Or is it deeper than that?

For example, does windows strip quotes from arguments before passing them to the underlying command, or do we have to strip them within the program?

bmaranville:
I think the short answer is that "%*" doesn't really work in PowerShell.  We can define a PowerShell function that passes through the args instead... I've already pushed one.

Issue #118, opened 2023-01-19
CI failing: updates for numpy 1.24 and scipy 1.10 needed


llimeht:
As well as updating CI to drop the unsupported Python 3.6 that was causing CI to fail (see #117), bumps also needs code updates for numpy 1.24 and scipy 1.10

### scipy 1.10

There seem to be changes around covariance matrix - I've not spotted the relevant information in the release notes, however. Reported failure is:

```
=================================== FAILURES ===================================
_____________________________________ test _____________________________________

    def test():
        """check entropy estimates from known distributions"""
        # entropy test is optional: don't test if sklearn is not installed
        try:
            import sklearn
        except ImportError:
            return
    
        # Smoke test - do all the methods run in 1-D and 10-D?
        _check_smoke(stats.norm(10, 8))
        if hasattr(stats, 'multivariate_normal'):
            _check_smoke(stats.multivariate_normal(cov=np.diag([1]*10)))
    
        D = stats.norm(10, 8)
        _check_entropy("N[100,8]", D, N=2000)
        _check_entropy("N[100,8]", D, N=12000)
        if hasattr(stats, 'multivariate_normal'):
            D = stats.multivariate_normal(cov=np.diag([1, 12**2, 0.2**2]))
            _check_entropy("MVN[1,12,0.2]", D)
            D = stats.multivariate_normal(cov=np.diag([1]*10))
            _check_entropy("MVN[1]*10", D, N=10000)
            # Make sure zero-width dimensions return H = -inf
            D = MVNSingular(cov=np.diag([1, 1, 0]))
>           _check_entropy("MVN[1,1,0]", D, N=10000)

bumps/dream/entropy.py:1074: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
bumps/dream/entropy.py:1010: in _check_entropy
    H = D.entropy()/LN2
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <bumps.dream.entropy.MVNSingular object at 0x7fb56e207100>, N = 10000

    def entropy(self, N=10000):
        with np.errstate(divide='ignore'):
>           return 0.5*log(np.linalg.det((2*pi*np.e)*self.dist.cov))
[...]

PR #120, opened 2023-01-27
Update code and tests for scipy 1.10 and numpy 1.24

llimeht:
This PR addresses the (in)compatibility of bumps with scipy 1.10 and numpy 1.24.

* Include the updates required for scipy 1.10 as discussed in #118 (thanks to Drew Parsons for the patch via the Debian packaging for bumps)
* Update np.float and np.int to used defined sizes for numpy 1.24 compatibility
* cherry-pick commit to drop Python 3.6 from CI; this is necessary to get *any* CI to pass at present as there python 3.6 is not available in the CI matrix any more. (Cherry-picked from #117; as noted there, CI for master currently fails and merging this commit is a prereq to get anything useful out of CI)

Closes: #118

### DIFF
--- a/bumps/dream/entropy.py
+++ b/bumps/dream/entropy.py
@@ -859,8 +859,14 @@ class MVNSingular:
         return self.dist.rvs(size=size)
 
     def entropy(self, N=10000):
+        # scipy 1.10 changed Covariance handling in stats, allow for that change
+        if hasattr(self.dist, 'cov'):
+            cov = self.dist.cov
+        else:
+            cov = self.dist.cov_object.covariance
+
         with np.errstate(divide='ignore'):
-            return 0.5*log(np.linalg.det((2*pi*np.e)*self.dist.cov))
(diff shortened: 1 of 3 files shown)


bumps/dream/entropy.py
```diff
@@ -859,8 +859,14 @@ def rvs(self, size=1):
         return self.dist.rvs(size=size)
 
     def entropy(self, N=10000):
+        # scipy 1.10 changed Covariance handling in stats, allow for that change
```
pkienzle, line 862: I try to mark compatibility changes with `# CRUFT $package < $version:` comments so that it is easier to clean up the code when we drop support for older versions of our dependencies. I will update the comment after merging.

Ideally I would perform compatibility checks at import time and monkey-patch old versions to the latest interface. I don't think I can do that here.

pkienzle: Looking at https://github.com/scipy/scipy/pull/17904 it looks like `dist.cov` will soon work again. I'll keep the cruft since the user might have a release version with it missing.

PR #127, opened 2023-06-13
Fall back to numba bounds checker in DREAM if compiled DLL not availalble

bmaranville:
In some cases (where calculation of the model nllf is very quick), the performance of the DREAM step can be a bottleneck.  A C version of the bounds checkers for parameters can be compiled as indicated in README.rst.

This PR adds numba.njit acceleration to the fallback `apply` methods in each bounds checker, providing performance comparable to the C versions, when numba is installed in the user's environment.

If numba is not installed or available, a non-accelerated version of the functions will still run (the njit decorator becomes a no-op passthrough)

Note that the previous behavior of the Bounds.apply(self, y) methods is changed, and they no longer return a value.  The return value of those functions was not used anywhere else in the bumps codebase.  In `partemp.py`, a locally-defined class ReflectBounds is already in use and is not affected by the change to bounds.ReflectBounds in this PR.

### DIFF
--- a/bumps/dream/bounds.py
+++ b/bumps/dream/bounds.py
@@ -70,18 +70,26 @@ class Bounds(object):
     c_interface = None # type: Callable[[int, int, Any, Any, Any], None]
     low = None # type: np.ndarray
     high = None # type: np.ndarray
-    def apply(self, x):
-        """Force x values within bounds"""
+
+    @staticmethod
+    def apply(minn, maxn, pop):
+        """Force pop (population) values within bounds"""
         raise NotImplementedError
 
-    def __call__(self, pop):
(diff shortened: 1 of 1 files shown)


bumps/dream/bounds.py
```diff
@@ -88,24 +92,24 @@ class ReflectBounds(Bounds):
     def __init__(self, low, high):
         self.low, self.high = [np.ascontiguousarray(v, 'd') for v in (low, high)]
 
-    def apply(self, y):
+    @staticmethod
+    @njit(cache=True)
+    def apply(minn, maxn, pop):
```
pkienzle, line 97: Need to change function signature in the Bounds class and update the tests to match the new interface.

PR #130, opened 2023-10-24
Add set_model()

mdoucet:
PR that goes along with the model builder PR in `refl1d`

### DIFF
--- a/bumps/webview/server/api.py
+++ b/bumps/webview/server/api.py
@@ -92,19 +92,28 @@ async def load_problem_file(pathlist: List[str], filename: str):
     state.problem.pathlist = pathlist
     await set_problem(problem, Path(*pathlist), filename)
 
-async def set_problem(problem: bumps.fitproblem.FitProblem, path: Optional[Path] = None, filename: str = ""):
+
+@register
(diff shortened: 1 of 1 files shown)


bmaranville:
There is a problem with this one, in that the new `api.set_model` does something very similar to what `api.set_problem` already does, and the two functions can be combined somewhat (maybe set_model calls set_problem after deserializing the problem?)
see: https://github.com/bumps/bumps/blob/64a8ce7cddccf63342f16e94827f697c77fd0156/bumps/webview/server/api.py#L95

Maybe we should add an argument on set_problem for not displaying the notification - you won't want to see that every time you update the model.

Also, `set_problem` calls `state.save`, writing out the problem to the HDF5 backing.  Do you want that to happen on every edit?  I'm not sure what the right answer is for that one.

bmaranville:
Now that I think about it, maybe we can rename your "set_model" to "update_problem" or something like that, where the implication is that it's an edit of an existing problem, and that we shouldn't call all the triggers/notifications for loading a new problem.

mdoucet:
@bmaranville Right. Good point. Have a look now.

Issue #91, opened 2022-03-17
update dependencies


pkienzle:
Make sure setup.py reflects current dependencies.

Drop 2.7 support and remove `six` as a dependency.

PR #136, opened 2024-02-25
remove dependency on six

a-detiste:
#91 

https://wiki.debian.org/Python3-six-removal

### DIFF
--- a/bumps/pmath.py
+++ b/bumps/pmath.py
@@ -3,7 +3,8 @@ Standard math functions for parameter expressions.
 """
 import math
 
-from six.moves import reduce, builtins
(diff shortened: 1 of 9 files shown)


bumps/pmath.py
```diff
@@ -3,7 +3,8 @@
 """
 import math
 
-from six.moves import reduce, builtins
+from functoons import reduce
```
pkienzle, line 6: functoons => functools

pkienzle (changes requested):
Minor typo in import, but otherwise I'm okay with it.

pkienzle (approved):

PR #138, opened 2024-02-29
Clear correlation_plot if nvars > MAX_CORR. Fixes #134

### LINKED ISSUE #134: CorrelationView not updated if number of parameters > 15
In bumps.gui.uncertainty_view CorrelationView, if a fit is run with 15 or fewer parameters, and then a subsequent fit is run using more parameters and the same persistent CorrelationView object, plots from the original fit remains. I understand, that when more than 15 parameters are fit, the correlation plots are not generated for various reasons, but not removing previous plots seems counter-intuitive.

pkienzle:
Allows you to set the number of correlations to show using `CorrelationView.MAX_CORR`, or `CorrelationView().MAX_CORR`.

If the number of correlations is exceeded it clears the figure and plots nothing.

### DIFF
--- a/bumps/gui/uncertainty_view.py
+++ b/bumps/gui/uncertainty_view.py
@@ -33,6 +33,12 @@ class UncertaintyView(PlotView):
 
 
 class CorrelationView(PlotView):
+    """
+    CorrelationView has a maximum number of correlations that it will show.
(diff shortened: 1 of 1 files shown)


krzywon (approved):
Exposing the limit looks like a good solution and seems to work.

PR #148, opened 2024-09-06
bumps webview update

pkienzle:


bumps/webview/server/state.py
```diff
@@ -0,0 +1,89 @@
+from collections import deque
```
pkienzle, line 1: Load/save state seems like it should be in the bumps directory rather than the webview server?

Can they or their equivalent be called from a jupyter notebook?

bumps/webview/server/varplot.py
```diff
@@ -0,0 +1,269 @@
+"""
+Build layout for histogram plots (plotly)
+"""
```
pkienzle, line 3: Can we replace the plotting functions in bumps.dream with these plotly versions rather than supporting two types?

pkienzle (commented):
Still need to review the logic of constraints, etc. as implemented in fitproblem.

There are clearly some bugs that are not currently being caught by tests.

Need more tests and documentation, particularly with the core parameters datastructures.

PR #151, opened 2024-09-10
numba implementation of dream stepper

pkienzle:
Minimally tested.

Need to check performance relative to compiled C and pure python versions.

### DIFF
--- a/bumps/dream/diffev.py
+++ b/bumps/dream/diffev.py
@@ -12,6 +12,7 @@ from numpy.linalg import norm
 from .util import draw, rng
 
 try:
+    #raise ImportError("skip numpy")
     from numba import njit, prange
     @njit(cache=True)
     def pchoice(choices, size=0, replace=True, p=None):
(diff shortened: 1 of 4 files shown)


bmaranville:
Seems like the C DLL will still be useful in extremely parallel environments - I don't see a good reason to get rid of it, unless we move the step logic into the parallelizer as well, as you suggested.  It's just a source file and a small set of compile instructions.

pkienzle:
It's hard to write a test to show that the compiled c and the numba are doing the same thing. Too many calls to the random number generator.

bumps/dream/diffev.py
```diff
@@ -109,22 +167,27 @@ def de_step(Nchain, pop, CR, max_pairs=2, eps=0.05,
         else:
             raise RuntimeError("Select failed...should never happen")
 
-        # If no step was specified (exceedingly unlikely!), then
```
bmaranville, line 112: Does this affect the stability if merged?  Will we see cases where no step was specified, and what will happen if we do?

bmaranville: After speaking with @pkienzle it seems like this won't be an issue

Issue #149, opened 2024-09-09
Publish package in pypi


jmborr:
When tag `vX.Y.Z` is created and pushed to the `master` branch, a dedicated GitHub Actions workflow will run with the following responsibilities:
- build a `noarch` pure Python wheel `bumps-X.Y.Z-py3-none-any.whl`
- for the following systems:
    + ubuntu-latest, py: 3.8, 3.9, 3.10, 3.11, 3.12
    + windows-latest, py: 3.9
    + macos-latest, py: 3.9
  install the wheel in a dedicated conda environment, then install necessary dependencies for testing. Run tests:
``` bash
pytest -v
python check_examples.py --chisq
python check_examples.py --chisq
```
(some of the test may make sense only for the `webview` branch)
- The python wheel will have the no entry points.
- Some of the [optional dependencies for webview](https://github.com/bumps/bumps/blob/master/pyproject.toml#L27) should be incorporated as package dependencies (which ones decided by trial an error, probably when running the tests).
### GitHub Workflow structure
Three worfklows:
    `Aw`. build the wheel, for all push events
    `wB` test with the wheel, for all push events
    `wC`. publish the wheel, for tag-push events with tag pattern `v*`
Workflow `wA` uploads the wheels as an artifact
- `wB` waits for `wA` ending with success, then downloads the wheel and
    + installs the wheel and runs the tests except those for webview
    + installs webview dependencies and runs the webview tests
- `wC` waits for `wB` to complete successfully

jmborr:
Completed

PR #154, opened 2024-09-10
Publish package in pypi

jmborr:
## Work Done
New GitHub action that will create, test, and publish a Python wheel in Pypi when a new tag with pattern `vX.Y.Z` or`vX.Y.Z.W` is pushed to the repo.
Notice that the wheel with have the version specified in `bumps.__init__.__version__` so this variable needs to be updated before pushing the tag!

References Issue #149


bmaranville (approved):
Looks good

PR #155, opened 2024-09-10
Coerce individual bounds

krzywon:
This coerces the bounds to +/- infinity if either bound is None.


bmaranville:
This looks fine to me

bmaranville (approved):

PR #156, opened 2024-09-11
get correct link to server in macos jupyter lab, which uses 127.0.0.1…

bmaranville:
… instead of localhost


bumps/webview/server/webserver.py
```diff
@@ -336,7 +336,7 @@ def get_server_url():
     # detect if running through Jupyter Hub
     if 'JUPYTERHUB_SERVICE_PREFIX' in os.environ:
         url = f"{os.environ['JUPYTERHUB_SERVICE_PREFIX']}/proxy/{port}/"
-    elif api.state.hostname == 'localhost': # local server
+    elif api.state.hostname == 'localhost' or api.state.hostname == '127.0.0.1': # local server
```
pkienzle, line 339: or `hostname in ("localhost", "127.0.0.1")` (ruff prefers double quoted strings).

pkienzle (approved):

PR #159, opened 2024-09-12
FIX: added support for Constant class

acaruana2009:
Add support for Constant class in `webview.api` in `params_to_list` function.

The lack of handling this class was causing the function to go into an infinite recursion.

### DIFF
--- a/bumps/webview/server/api.py
+++ b/bumps/webview/server/api.py
@@ -16,7 +16,7 @@ import uuid
 
 from bumps.fitters import DreamFit, LevenbergMarquardtFit, SimplexFit, DEFit, MPFit, BFGSFit, FitDriver, fit, nllf_scale, format_uncertainty
 from bumps.mapper import MPMapper
-from bumps.parameter import Parameter, Variable, unique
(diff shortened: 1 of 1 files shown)


bumps/webview/server/api.py
```diff
@@ -925,20 +925,28 @@ def params_to_list(params, lookup=None, pathlist=None, links=None) -> List[Param
                 new_pathlist.append("")
             new_pathlist[-1] = f"{new_pathlist[-1]}[{i:d}]"
             params_to_list(v, lookup=lookup, pathlist=new_pathlist)
-    elif isinstance(params, Parameter):
+    elif isinstance(params, Parameter) or isinstance(params, Constant):
         path = ".".join(pathlist)
         existing = lookup.get(params.id, None)
```
pkienzle, line 930: Not sure that `parameter.id` is unique. Maybe we should use `id(parameter)` instead so we know that it is pointing to the same object?

I realize this is not in the scope of this PR. Feel free to turn it into an issue instead.

bmaranville (approved):

PR #160, opened 2024-09-12
Custom plots

hoogerheide:
Adds custom plotting hooks to `Fitness`-type objects (in this case, `bumps.Curve` to be extended to `refl1d.ExperimentBase`).

Custom plotting functions are defined by registering them to a model with a descriptive string using `register_webview_plot`.

When the plot is requested, a copy of the model is refreshed (with `.update()` and `.nllf()`, and the arguments `model`, `problem`, and `state` passed through. The function returns a `CustomWebviewPlot(TypedDict)` object with fields `fig_type` and `plotdata`, following the current naming conventions.

Currently `plotly`, `matplotlib`, and `error` are allowed types; `error` displays the traceback of the plotting function to enable real-time debugging.

Subclasses of `Fitness`-type objects can automatically register any plotting functions intrinsic to that class, allowing the registration process to be compatible with serialization of these classes.

### DIFF
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -359,9 +359,9 @@ class Curve(object):
         self._plot_callbacks.update({plot_title: dict(func=plot_function,
                                                       change_with=change_with)})
 
-    def create_webview_plot(self, title, problem, state):
+    def create_webview_plot(self, title, problem, state, nshown):
 
-        return self._plot_callbacks[title]['func'](self, problem, state)
+        return self._plot_callbacks[title]['func'](self, problem, state, nshown)
 
 def _plot_resids(x, resid, colors, labels, view):
     import pylab
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -43,6 +43,7 @@ __all__ = ["Curve", "PoissonCurve", "plot_err"]
 
 import inspect
 import warnings
+from typing import Callable
 
 import numpy as np
 from numpy import log, pi, sqrt
@@ -234,6 +235,7 @@ class Curve(object):
         self._state = state
         self._plot = plot
         self._cached_theory = None
+        self._plot_callbacks = {}
 
     def update(self):
         self._cached_theory = None
@@ -345,6 +347,19 @@ class Curve(object):
 
             pylab.subplot2grid((plot_ratio, 1), (plot_ratio-1, 0), sharex=h)
             _plot_resids(x, resid, colors=colors, labels=labels, view=view)
+
+    def register_webview_plot(self, plot_title: str, plot_function: Callable):
+        # Plot function syntax: f(model, problem, state)
+        
+        self._plot_callbacks.update({plot_title: plot_function})
+
+    def get_plot_titles(self):
+
+        return list(self._plot_callbacks.keys())
+    
+    def get_plot(self, title, problem, state):
+
+        return self._plot_callbacks[title](self, problem, state)
 
 def _plot_resids(x, resid, colors, labels, view):
     import pylab
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -361,7 +361,7 @@ class Curve(object):
 
     def create_webview_plot(self, title, problem, state):
 
-        return self._plot_callbacks[title]['func'](deepcopy(self), problem, state)
+        return self._plot_callbacks[title]['func'](self, problem, state)
 
 def _plot_resids(x, resid, colors, labels, view):
     import pylab
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -43,7 +43,7 @@ __all__ = ["Curve", "PoissonCurve", "plot_err"]
 
 import inspect
 import warnings
-from typing import Callable
+from typing import Callable, Literal
 
 import numpy as np
 from numpy import log, pi, sqrt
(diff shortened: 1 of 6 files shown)


hoogerheide:
This is the first stage of this PR. The next stage is to enable Custom Uncertainty plots that refresh only when the uncertainty state is updated.

I am unclear on whether the uncertainty plot architecture (which displays the spinner when the plot is refreshing) could just be used for immediately responsive plots, or how to discriminate between these without making a separate set of hooks. Thoughts welcome.

bmaranville:
The spinner in those plots is on a delay - if your plot renders faster than a specified timeout, the spinner is never shown, and it just shows your plot.

hoogerheide:
Why don't all the plots use it then?

bmaranville:
Sure - I think it's just extra code that isn't needed for the "faster" plotters, so I never patched it back in.

bumps/curve.py
```diff
@@ -346,6 +348,19 @@ def plot(self, view=None):
             pylab.subplot2grid((plot_ratio, 1), (plot_ratio-1, 0), sharex=h)
             _plot_resids(x, resid, colors=colors, labels=labels, view=view)
 
+    def register_webview_plot(self, plot_title: str, plot_function: Callable):
+        # Plot function syntax: f(model, problem, state)
+        
+        self._plot_callbacks.update({plot_title: plot_function})
+
+    def get_plot_titles(self):
```
pkienzle, line 356: Python generally does not use get/set in names. For example, `self._plot_callbacks.keys()` for get. For the setter it would use the same name, but with an argument for the new value (doesn't apply for keys).

For "cheap" queries use the `@property` decorator so it looks like an attribute.

bumps/webview/server/api.py
```diff
@@ -551,6 +552,49 @@ async def get_model():
...
+@register
+async def get_custom_plot_names():
+    if state.problem is None or state.problem.fitProblem is None:
+        return None
+    fitProblem = state.problem.fitProblem
+    output: List[Dict[int, str]] = []
+    for model_index, model in enumerate(fitProblem.models):
+        output += [dict(model_index=model_index,
+                        name=title)
+                        for title in model.get_plot_titles()]
```
pkienzle, line 564: General formatting guide: indents only increase by four characters.  Rather than trying to maintain alignment with the previous open, break after the `[` putting dict() on a line by itself, and detent `for` to the same level.

Ruff will reformat this.

bumps/curve.py
```diff
@@ -346,6 +348,19 @@ def plot(self, view=None):
             pylab.subplot2grid((plot_ratio, 1), (plot_ratio-1, 0), sharex=h)
             _plot_resids(x, resid, colors=colors, labels=labels, view=view)
 
+    def register_webview_plot(self, plot_title: str, plot_function: Callable):
+        # Plot function syntax: f(model, problem, state)
+        
+        self._plot_callbacks.update({plot_title: plot_function})
```
bmaranville, line 354: as I have seen it in use, `callback` denotes a function or pointer to a function that is passed into another "receiver" function, typically to be called with the output of the receiver function when that function has completed all other execution, or passed on to the next function in a callback chain.  I think it would be more accurate to call these `plot_functions`, as they are selected and then directly called, not passed on or deferred.

bumps/webview/server/api.py
hoogerheide: Agreed that the for should be dedented, but not about the four character indents. See the first example here in the PEP8 standard: https://peps.python.org/pep-0008/#indentation

bumps/curve.py
bmaranville: I stand corrected - `callback` is a term that can be applied to any function pointer, while `callback parameter` refers more closely to the usage I was describing.

bumps/curve.py
hoogerheide: Resolved by https://github.com/bumps/bumps/pull/160/commits/dd6abe6bb4581292473df4770368fa29bac0c255

hoogerheide:
Please check out the latest commit, which also adds a custom uncertainty panel.

bumps/webview/server/api.py
```diff
@@ -570,7 +570,7 @@ async def create_custom_plot(model_index: int, plot_title: str) -> CustomWebview
     if state.problem is None or state.problem.fitProblem is None:
         return None
     fitProblem = deepcopy(state.problem.fitProblem)
-    uncertainty_state = state.fitting.uncertainty_state
+    uncertainty_state = deepcopy(state.fitting.uncertainty_state)
```
bmaranville, line 573: The uncertainty state is potentially very large - why copy it?  This risks users running out of memory

hoogerheide: Good point, and the state doesn't get modified at all. It might be worth discussing whether it makes sense to send the entire state, or just a draw from the state -- do we ever use anything besides the points and logp?

bumps/webview/server/custom_plot.py
```diff
@@ -0,0 +1,26 @@
+from typing import Literal, TypedDict, Any
+
+class CustomWebviewPlot(TypedDict):
+    fig_type: Literal['plotly'] | Literal ['matplotlib'] | Literal['error'] = 'plotly'
+    plotdata: Any
+
+def process_custom_plot(plot_item: CustomWebviewPlot) -> CustomWebviewPlot:
```
bmaranville, line 7: This function seems unnecessary - we can just require that plot functions return a valid dict as plot_data

hoogerheide: This would require mpld3 imports into users' scripts. While this may not be the usual use case, do we want to require this?

bmaranville: Ah, I see. It would certainly make it more complicated for users to add plots in a script.  I don't think of that as a very likely use case, but it would get a lot less likely if we made it hard.

bumps/webview/server/api.py
hoogerheide: Removed in https://github.com/bumps/bumps/pull/160/commits/0b076df0c7251ce2f8b86b6c039f6c65ca7b49c8.

bmaranville:
Superseded by #167

PR #163, opened 2024-09-18
adding mystic to packages to install

bmaranville:


bmaranville:
This seems to have worked - the mystic subdirectory is in the wheel file.

PR #167, opened 2024-10-01
Custom plots rebase

bmaranville:
To replace #160

### DIFF
--- a/bumps/webview/server/custom_plot.py
+++ b/bumps/webview/server/custom_plot.py
@@ -1,7 +1,7 @@
 from typing import Literal, TypedDict, Any
 
 class CustomWebviewPlot(TypedDict):
-    fig_type: Literal['plotly'] | Literal ['matplotlib'] | Literal['error'] = 'plotly'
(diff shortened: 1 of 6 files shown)


bumps/webview/server/custom_plot.py
```diff
@@ -0,0 +1,26 @@
...
+        figdict = plot_data.to_dict()
+    elif figtype == 'matplotlib':
+        import mpld3
+        figdict = mpld3.fig_to_dict(plot_data)
+    elif figtype == 'error':
+        figdict = plot_data
+    else:
+        figdict = dict(error=f'unrecognized plot type {figtype}')
+    
+    del plot_data # is this necessary? Does plot_item['plot_data'] also need to be deleted?
```
bmaranville, line 22: No, not really - `plot_data` and even the `plot_item` dict passed in to this function will be garbage collected when the function goes out of scope (if there are no other references to them), since you're creating a new CustomWebviewPlot at the end of the function.

PR #168, opened 2024-10-02
Trace plot update

hoogerheide:
Update to the parameter trace plot. Major changes:
1. Plot is now rendered in plotly
2. Any fittable parameter can be selected for inspection.
3. Colors has been moved from refl1d to bumps

One open issue: at the moment the parameter trace plot is using `uncertainty_state`, and it updates with the `"updated_uncertainty"` signal. It is not clear to me if the information in this plot can be updated more regularly; a more natural update signal would be `"updated_convergence"`, but of course at the moment the uncertainty state isn't updated with the convergence updates so this doesn't really make sense.

### DIFF
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -359,9 +359,9 @@ class Curve(object):
         self._plot_callbacks.update({plot_title: dict(func=plot_function,
                                                       change_with=change_with)})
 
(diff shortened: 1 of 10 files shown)


hoogerheide:
History has gotten somewhat complicated, will try clean PR after rebase

PR #170, opened 2024-10-03
Trace plot update

hoogerheide:
Fresh version of https://github.com/bumps/bumps/pull/168

Update to the parameter trace plot. Major changes:

1. Plot is now rendered in plotly
2. Any fittable parameter can be selected for inspection.
3. Colors has been moved from refl1d to bumps

An open question:
At the moment the parameter trace plot is using uncertainty_state, and it updates with the "updated_uncertainty" signal. It is not clear to me if the information in this plot can be updated more regularly; a more natural update signal would be "updated_convergence", but of course at the moment the uncertainty state isn't updated with the convergence updates so this doesn't really make sense.

### DIFF
--- a/bumps/webview/server/api.py
+++ b/bumps/webview/server/api.py
@@ -758,17 +758,12 @@ async def get_parameter_labels():
 
     if state.problem is None or state.problem.fitProblem is None:
         return None
-    fitProblem = state.problem.fitProblem
-    if fitProblem is not None:
-        return to_json_compatible_dict(fitProblem.labels())
-    else:
-        return None
+    return to_json_compatible_dict(state.problem.fitProblem.labels())
 
 @register
 async def get_parameter_trace_plot(var: int):
     
     uncertainty_state = state.fitting.uncertainty_state
-    logger.info(f'Got parameter_trace plot request with index {var}')
     if uncertainty_state is not None:
         import time
 
@@ -782,7 +777,7 @@ async def get_parameter_trace_plot(var: int):
         start = int((1-portion)*len(draw)) if portion else 0
         genid = np.arange(uncertainty_state.generation-len(draw)+start, uncertainty_state.generation)+1
         fig = plot_trace(genid*uncertainty_state.thinning,
-             np.squeeze(points[start:, uncertainty_state._good_chains, var]).T, label=label, alpha=0.2)
+             np.squeeze(points[start:, uncertainty_state._good_chains, var]).T, label=label, alpha=0.4)
 
         logger.info(f"time to render but not serialize... {time.time() - start_time}")
         dfig = fig.to_dict()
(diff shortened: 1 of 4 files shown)


bmaranville:
The "updated_uncertainty" signal is sent every time the fit thread sends a new copy of the uncertainty state to the server, so you can't really access it faster than that from the client.

On the other hand, it has been planned for quite some time to expose the uncertainty update time as a user parameter in the fit options panel, it's just yet another thing we haven't gotten to yet!

If you could set that to something shorter than the current 300 seconds, I think that might address your issue.

hoogerheide:
This is probably fine. I guess logp is being sent back in "real time", which would be the information used in the convergence plot. I can't really imagine anybody sitting and watching the trace plot -- it's probably not more useful than setting the update uncertainty timer to 60 seconds or so, which is what you suggested. Let's not worry about this and just leave it as updating with the uncertainty.

I *do* like the idea of allowing the uncertainty state to be updated, at a different rate from the autosave interval, during fits.

bumps/webview/server/api.py
```diff
@@ -748,40 +749,41 @@ async def get_model_uncertainty_plot():
...
         label = uncertainty_state.labels[var]
         start = int((1-portion)*len(draw)) if portion else 0
         genid = np.arange(uncertainty_state.generation-len(draw)+start, uncertainty_state.generation)+1
-        axes.plot(genid*uncertainty_state.thinning,
-             np.squeeze(points[start:, uncertainty_state._good_chains, var]))
-        axes.set_xlabel('Generation number')
-        axes.set_ylabel(label)
-        fig.canvas.draw()
+        fig = plot_trace(genid*uncertainty_state.thinning,
+             np.squeeze(points[start:, uncertainty_state._good_chains, var]).T, label=label, alpha=0.4)
```
bmaranville, line 780: What is `np.squeeze` doing in this context?

bumps/webview/server/api.py
```diff
@@ -748,40 +749,41 @@ async def get_model_uncertainty_plot():
...
+    # get_parameters. Uses fitProblem because uncertainty state might not exist and
+    # list of parameters should be updated on model_loaded.
+    # Should probably be able to call these parameters by ID.
+
+    if state.problem is None or state.problem.fitProblem is None:
+        return None
+    return to_json_compatible_dict(state.problem.fitProblem.labels())
+
+@register
+async def get_parameter_trace_plot(var: int):
```
bmaranville, line 764: Would it make sense to define a function `get_parameter_trace_plot(uncertainty_state: MCMCDraw, var_index: int=0)` in `traceplot.py`, instead of defining it here, since you're already splitting off a smaller part of the plot implementation there?

bmaranville (changes requested):
Looks good overall, but I think it might be a lot cleaner to move the whole plot implementation to traceplot.py (taking the uncertainty state as an argument)

bumps/webview/server/api.py
hoogerheide: It removes the extra dimension from `points`. The last dimension of `points` has length Nvar, but this dimension persists when selecting the parameter with index `var`. `np.squeeze` gets rid of the extra dimension.

bumps/webview/server/api.py
bmaranville: That's fair - I agree with separating the plotting-library-specific code off.

bmaranville (approved):

bmaranville:
Is the colors change already merged in bumps?

PR #171, opened 2024-10-03
Custom plot table

hoogerheide:
Adds webview support for static tabular data provided as CSV.

Currently only used as a custom plot option, but could also be used to create standard tabs containing tabular data (most immediately a parameter statistics tab showing 68% and 95% CIs on the fit parameters).

This implementation is extremely lightweight and extensible:
1. The base python CSV parser is used to generate header and row data
2. I made a new Vue component to render the table from the parsed data and allow the user to download the raw data (note that the raw data are included along with the parsed data to avoid an extraneous server call; I imagine we'll mostly be dealing with small tables so the duplication makes sense).

Screenshot (formatting / contents controlled by source, not the webview renderer):

image

### DIFF
--- a/bumps/webview/client/src/components/CSVTable.vue
+++ b/bumps/webview/client/src/components/CSVTable.vue
@@ -29,14 +29,14 @@ async function download_csv() {
         </div>
         <table class="table">
             <thead class="border-bottom py-1 sticky-top text-white bg-secondary">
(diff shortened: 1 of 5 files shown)


bmaranville (approved):
This looks good - did you test to see that the CSV download works?

PR #172, opened 2024-10-16
Remove client CDN, force developers to build

bmaranville:
The webview server currently checks to see if a client is built in `bumps/webview/client/dist/<version from package.json>/` at startup, and if no matching client is found, it instead shows a templated index.html file that points to a CDN where a compiled client for that version might be found.

We had been regularly publishing the built clients to npmjs, which then automatically made them available on jsdelivr.com

`CDN_TEMPLATE = "https://cdn.jsdelivr.net/npm/bumps-webview-client@{client_version}/dist/{client_version}"`

Now, we are regularly publishing wheels for bumps through pypi which include a prebuilt client in the `bumps/webview/client/dist/` folder, and this fallback is no longer needed.  We are also providing a convenience script so that developers can build an up-to-date client on demand with 
```sh
python -m bumps.webview.build_client
```

In this PR, the CDN fallback is removed from the server and it is assumed that the client exists.  If no client is found, and the index page is requested, a custom 404 page is shown with instructions to run the build command above.

The client is no longer built into the subfolder with a version number, and the default `vite` behavior is used, which places the client files directly in the `dist` folder.  The version number in `package.json` will no longer be incremented, and the effective version of the client source files will be the version of the python package.

In the future we might try to update the `package.json` using the `versioningit` infrastructure, at the same time the python version is updated when a new tag appears.


bmaranville:
Closing in favor of #200

PR #176, opened 2024-10-23
Webview cli fit options

bmaranville:
Add an cli flag for setting fit options when starting the webview server.

Example usage:
`bumps-webview --fit=dream --fit_options init=eps steps=2000 --start --exit --write_store=output.h5 mymodel.py`,

This would

- load the model `mymodel.py` from the current folder
- set `dream` as the fitter (from `--fit`) with the default dream settings, but overriding
   - `steps` and 
   - `init` (from `--fit_options init=eps steps=2000`)
- define `output.h5` as the path to write a session file (from `--write_store`)
   - this will overwrite any file called `output.h5` in the current directory
   - to append to an existing session file, use `--store` instead, to set the `read_store` and `write_store` at the same time, but be aware that the problem file `mymodel.py` will be loaded, replacing any active model in the session file.  To use the model in the session, just omit the problem file name in the above call. 
- start the fit (from `--start`)
- terminate the server when the fit is complete and the session file has been written. (from `--exit`)


bumps/webview/server/webserver.py
```diff
@@ -136,11 +139,12 @@ def get_commandline_options(arg_defaults: Optional[Dict]=None):
     parser.add_argument('--exit', action='store_true', help='end process when fit complete (fit results lost unless write_store is specified)')
     parser.add_argument('--serializer', default=OPTIONS_CLASS.serializer, type=str, choices=["pickle", "dill", "dataclass"], help='strategy for serializing problem, will use value from store if it has already been defined')
     parser.add_argument('--trace', action='store_true', help='enable memory tracing (prints after every uncertainty update in dream)')
-    parser.add_argument('--parallel', default=0, type=int, help='run fit using multiprocessing for parallelism; use --parallel=0 for all cpus')
+    parser.add_argument('--parallel', default=0, type=int, help='run fit using multiprocessing for parallelism; default is --parallel=0 for all cpus')
```
acaruana2009, line 142: Will this act as an override for the GUI? In the past when I have tried to use this to limit CPUs it has ended up still using all CPUs in the GUI anyway.

bumps/webview/server/fit_options.py
acaruana2009, line 1: I am assuming this is mostly being pulled from the old CLI file? I am fine with this if it declutters cli.py. Just want to make sure we are not duplicating code here.

bumps/webview/server/webserver.py
```diff
@@ -136,11 +139,12 @@ def get_commandline_options(arg_defaults: Optional[Dict]=None):
...
     parser.add_argument('--path', default=None, type=str, help='set initial path for save and load dialogs')
     parser.add_argument('--no_auto_history', action='store_true', help='disable auto-appending problem state to history on load and at fit end')
     parser.add_argument('--convergence_heartbeat', action='store_true', help='enable convergence heartbeat for jupyter kernel (keeps kernel alive during fit)')
     parser.add_argument('--use_persistent_path', action='store_true', help='save most recently used path to disk for persistence between sessions')
+    parser.add_argument('--fit_options', nargs='*', type=str, help='fit options as key=value pairs')
```
acaruana2009, line 147: So this would be a slight departure from specifying all of the fit options as separate args? I.e. in the past it would have been a case of calling -- for every arg, and having the fitter work this out - i.e. if the fitter didn't have access to the argument it just ignored it. I don't mind this change in syntax but feels like it is something else to remember when using a cli call.

acaruana2009 (commented):
In general, I am happy for this to be merged, but might want a little bit of a discussion about how the fit options syntax has been altered. If we can avoid things that make it more tricky to remember how to use that is preferred. That said if there is a good coding/maintenance argument for going this way then I am happy.

bumps/webview/server/webserver.py
bmaranville: In the short term, I think we should make the CLI arguments for webview compatible with the existing CLI, since they are likely to be merged into a single interface in the future.

acaruana2009: I think that makes sense.

acaruana2009:
We talked about before, making some option to be able to use model arguments - currently this is possible with the cli. Are we aiming to add that here?

bumps/webview/server/fit_options.py
```diff
@@ -0,0 +1,64 @@
...
+    "radius": ["Simplex radius", "float"],
+    "trim": ["Burn-in trim", "boolean"],
+    "outliers": ["Outliers", ["none", "iqr", "grubbs", "mahal"]],
+}
+
+
+def parse_fit_options(fitter_id: str, fit_options: Optional[List[str]] = None) -> Dict:
+    if fitter_id not in api.FITTER_DEFAULTS:
+        raise ValueError(f"invalid fitter: {fitter_id}")
+    fitter_settings: Dict = api.FITTER_DEFAULTS[fitter_id]["settings"]
```
pkienzle, line 30: Do you want a copy here? Right now you are changing the defaults in the fitter, which is probably required for the current version since you are not passing the updated options to the GUI. I would prefer that parsing were separate from updates to global state.

The option updates only applies to the fitter on the command line not all the fitters.  That's okay if specifying --fit and --fit-options implicitly starts the fitter and perhaps exits, but if the primary goal is to tweak the defaults in the GUI if feels like this would be better handled by a preferences file.

I guess the fit options for each fit are saved in the session file? I wonder if clicking on an old fit should restore the fitter controls to the values for that fit.

bumps/webview/server/webserver.py
```diff
@@ -366,7 +384,7 @@ async def load_problem(App=None):
 
         async def start_fit(App=None):
             if api.state.problem is not None:
-                await api.start_fit_thread(fitter_id, fitter_settings["settings"], options.exit)
+                await api.start_fit_thread(fitter_id, fitter_settings, options.exit)
```
pkienzle, line 387: I may have broken things with this change. The `apply_fit_options` function was returning the settings dictionary so it seems like settings was not going to be available within this object.

pkienzle (approved):
As mentioned elsewhere, cli and webview should use the same options processing. I did some checks on argparser and it seems to handle the way I use the command line. We could pull the features from cli into webview and eventually drop cli.

bumps/webview/server/webserver.py
bmaranville, line 1: This is broken at the moment - can't run.  Should we try to fix it, or revert it, now that it might not be combined with cli.py (I don't think it should be combined with cli.py, necessarily)

bmaranville:
Ok, I think it's working at the moment (starting a fit from the cli, and starting a fit from the webview)

pkienzle:
Experiencing frequent HDF errors on windows when running the following command repeatedly:
```sh
python -m bumps.webview.server.cli doc\examples\curvefit\curve.py --store=T1.hdf
```

```python
Traceback (most recent call last):
  File "<frozen runpy>", line 198, in _run_module_as_main
  File "<frozen runpy>", line 88, in _run_code
  File "...\bumps\webview\server\cli.py", line 621, in <module>
    main()
  File "...\bumps\webview\server\cli.py", line 617, in main
    run_batch_fit(options)
  File "...\bumps\webview\server\cli.py", line 570, in run_batch_fit
    asyncio.run(_run_operations(on_startup, on_complete))
  File "...\asyncio\runners.py", line 194, in run
    return runner.run(main)
           ^^^^^^^^^^^^^^^^
  File "...\asyncio\runners.py", line 118, in run
    return self._loop.run_until_complete(task)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\asyncio\base_events.py", line 687, in run_until_complete
    return future.result()
           ^^^^^^^^^^^^^^^
  File "...\bumps\webview\server\cli.py", line 516, in _run_operations
    await step(None)
  File "...\bumps\webview\server\cli.py", line 396, in load_problem
    await api.load_problem_file(pathlist, filename, args=options.model_args)
  File "...\bumps\webview\server\api.py", line 171, in load_problem_file
    await set_problem(problem, Path(*pathlist), filename)
  File "...\bumps\webview\server\api.py", line 206, in set_problem
    state.autosave()
  File "...\bumps\webview\server\state_hdf5_backed.py", line 468, in autosave
    self.save()
  File "...\bumps\webview\server\state_hdf5_backed.py", line 475, in save
    self.write_session_file(full_path)
  File "...\bumps\webview\server\state_hdf5_backed.py", line 486, in write_session_file
    self.history.write(root_group)
  File "...\bumps\webview\server\state_hdf5_backed.py", line 230, in write
    problem.write(item_group)
  File "...\bumps\webview\server\state_hdf5_backed.py", line 194, in write
    write_fitproblem(group, "fitProblem", self.fitProblem, self.serializer)
  File "...\bumps\webview\server\state_hdf5_backed.py", line 132, in write_fitproblem
    dset = write_bytes_data(group, name, serialized)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\bumps\webview\server\state_hdf5_backed.py", line 98, in write_bytes_data
    return group.create_dataset(name, data=np.void(saved_data), compression=COMPRESSION)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\h5py\_hl\group.py", line 183, in create_dataset
    dsid = dataset.make_new_dset(group, shape, dtype, data, name, **kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "...\h5py\_hl\dataset.py", line 163, in make_new_dset
    dset_id = h5d.create(parent.id, name, tid, sid, dcpl=dcpl, dapl=dapl)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "h5py\\_objects.pyx", line 54, in h5py._objects.with_phil.wrapper
  File "h5py\\_objects.pyx", line 55, in h5py._objects.with_phil.wrapper
  File "h5py\\h5d.pyx", line 137, in h5py.h5d.create
ValueError: Unable to synchronously create dataset (name already exists)
```

bmaranville:
I think the HDF error you're seeing might come from a name collision of the history items if two history items have the same timestamp (to the millisecond).  We can address this in a separate PR.

bumps/webview/server/api.py
```diff
@@ -1191,8 +1239,13 @@ async def shutdown():
     logger.info("killing...")
     await stop_fit()
```
bmaranville, line 1240: This no longer waits for the fit to actually be done... all the awaits were removed from stop_fit

bumps/webview/server/state_hdf5_backed.py
```diff
@@ -659,16 +658,16 @@ async def get(self, name):
 
     def write(self, parent: "Group"):
         group = parent.require_group("shared")
-        for field in fields(self):
-            if not field.name in self._not_reloaded:
```
bmaranville, line 657: does ruff complain about this order?  `not <boolean expression>` always makes more sense to me than `<boolean expression with not somewhere in it>`, because the first is a straightforward negation of a thing that evaluates to boolean.

pkienzle:
Checking if comment to a PR triggers a pull_request action

bumps/webview/server/state_hdf5_backed.py
pkienzle: Yes, I was removing lint warnings that were appearing in VS Code

bumps/webview/server/webserver.py
pkienzle: What is broken? I replaced `field` with `f` so it wouldn't shadow the toplevel import of field from dataclasses.

PR #177, opened 2024-10-23
Add extension to session file if none provided

glass-ships:
Appends `session.h5` if no extension is provided when saving a session. 

Additionally includes:
- In `SessionMenu.vue` cast `$event.target` as HTMLInputElement
- Remove `defineProps` import, as it's now a compiler method


bumps/webview/client/src/asyncSocket.ts
```diff
@@ -1,26 +1,26 @@
-import { Socket } from 'socket.io-client';
+import { Socket } from "socket.io-client";
 
-declare module 'socket.io-client' {
+declare module "socket.io-client" {
```
bmaranville, line 3: There are a bunch of format changes in here unrelated to the logical changes - is there a specific formatting rule being applied?  Are the tooling settings updated for ruff or something so that these changes are reproducible?

glass-ships: Apologies, this is `prettier`, ruff is exclusive to python. 

We could add prettier and a "lint" script in the package.json, so one could run `npm run lint` and it would apply linting across the package, in a similar way to ruff

I just have a habit of constantly pressing my VSCode keyboard shortcut to format a document. 

That's probably something that would be better done in a separate PR though, maybe something that sets up pre-commit and ruff in a similar way to what we did with refl1d?

bumps/webview/client/src/components/SessionMenu.vue
```diff
@@ -167,7 +167,7 @@ async function unsetOutputFile() {
           <label for="autosaveIntervalInput" class="col col-form-label col-form-label-sm">Interval (s)</label>
           <div class="col-auto">
             <input type="number" class="form-control form-control" id="autosaveIntervalInput"
-            :value="autosave_session_interval" @change="set_interval($event.target.valueAsNumber)">
+            :value="autosave_session_interval" @change="set_interval((($event.target as HTMLInputElement)?.valueAsNumber) ?? 0)">
```
bmaranville, line 170: Seems like a good idea to cast target to HTMLInputElement, since it is one, but the `@change` handler cannot be fired before the element exists, and so it always has the `valueAsNumber` property (no reason to guard that access), and also it will never be undefined (so no need for the `??` either)... we should probably be checking for `NaN` instead, since that will really be wrong and might cause errors.

(The `??` operator does not filter `NaN`)

bumps/webview/client/src/components/FileBrowser.vue
```diff
@@ -124,17 +121,39 @@ async function setPath(new_pathlist?: string[]) {
...
-  const scale = (bytes > 0) ? Math.min(5, Math.floor(Math.log(bytes) / LOG_1024)) : 0;
+  const scale = bytes > 0 ? Math.min(5, Math.floor(Math.log(bytes) / LOG_1024)) : 0;
   const prefix = PREFIXES[scale];
   const reduced = bytes / SCALES[scale];
-  const extra_digits = (scale == 0) ? 0 : 1;
+  const extra_digits = scale == 0 ? 0 : 1;
   return `${reduced.toFixed(extra_digits)} ${prefix}B`;
 }
 
+/** TODO: what if chosenFile has a / in it? */
```
bmaranville, line 133: If we do anything about slashes, we should probably make them illegal here, since there's no platform-independent way to deal with them.  We could raise a modal alert() that indicates that you can't have slashes in a filename, and **not** close the dialog or take any other action (return from chooseFile after alert)

bumps/webview/client/src/components/FileBrowser.vue
```diff
@@ -124,17 +121,39 @@ async function setPath(new_pathlist?: string[]) {
...
+    // filename has no dot
+    filename = `${input}.session.h5`;
+  } else if (input.endsWith(".h5") || input.endsWith(".hdf5")) {
+    // filename has a valid extension
+    if (input.includes("session")) {
+      // filename has "session"
+      filename = input;
+    } else {
+      // add "session" before the extension
+      filename = `${input.split(".").slice(0, -1).join(".")}.session.${input.split(".").pop()}`;
```
bmaranville, line 149: inserting an extension before the end is a pretty unusual thing for an application to do - I think it would be unexpected and confusing.  

I think it's common to add an extension if none exists, but leave the filename alone if the user explicitly chosen an extension, even if it doesn't match the desired extension.  This is how people can override the default terminal extension, should they so desire.

bumps/webview/client/src/components/FileBrowser.vue
glass-ships: I'm pretty sure pathlib is platform independent, in that: 
`Path("some_dir") / "subdir"` resolves to the appropriate string depending on windows/unix

so we could just split `"inputdir/asubdir"` on `"/"` and push that to `pathlist` which is already a list of strings that gets passed to the python API, which turns it into `Path(*pathlist)`

bmaranville: True, but the platform ambiguity is still there - a user on Windows might want to use backslashes to denote a subfolder - do we split on both types of slashes simultaneously?  How often are users really going to want to do this?  Would users expect that a subfolder tree be created if it doesn't already exist? (that is a slightly atypical UX)

I guess I'm worried about adding complexity and non-standard UX, for a feature that might not get used very often.

bmaranville: Though we don't currently have an "add subfolder" feature in the file browser... and this would address that.

krzywon: pathlib Path objects handle the platform ambiguity for you. There should be no need to split input paths.

glass-ships: i suppose for now we could just disallow slash or backslash and address it in a future PR. 

@krzywon it does, however: 
```python
from pathlib import Path


def test(input):
    res = Path(*input) if input else Path("use_base")
    return res


if __name__ == "__main__":
    print(test(None))                           # use_base
    print(test(0))                              # use_base
    print(test([]))                             # use_base
    print(test(["test", "with/slash"]))         # test/with/slash
    print(test(["test", "with\\backslash"]))    # test/with\backslash
```

as an aside, it can definitely be beneficial to consider third party components where possible, and something like a file browser is a good use case for that, as someone has very likely already designed something that fulfills the needs here.

glass-ships:
Comments addressed - ready for re-review

bumps/webview/client/src/components/FileBrowser.vue
bmaranville: I agree completely!  I looked around some but didn't find an off-the-shelf solution, but would be interested in finding one.

glass-ships: i'll do some digging around for a potential future update

bmaranville (approved):

bmaranville:
There seems to be a lot of commit history with this - when I pull this branch and try to rebase it on `master`, it shows that it is trying to apply all the formatting changes and deletes etc. then trying to revert them, and failing because of a conflict on `api.py`.  Is it possible to get a rebased version of this with only the changes reflected in the final state?

glass-ships:
are we able to just complete the merge into master in the git UI and squash commits?

bmaranville:
That seems like a good idea.

bmaranville:
Ack!  This is breaking things... the extension is being added for all uses of the file browser, not just when being used from the session menu.  I should have caught that.  Need a quick fix.

glass-ships:
Where else is the FileBrowser used?

bmaranville:
Everywhere we interact with files!

PR #183, opened 2024-11-08
separate the generation of error points to calculate from the calculation call

bmaranville:
In errplot, the function `calculate_errors_from_state` does 2 distinct things:

- generates a list of points at which to calculate
- calls the `calc_errors` function on those points, with the fitProblem (which then calls `fitplugin.calc_errors` with the problem and points)

Here we separate this into two functions, one of which generates the list of points, and the other of which calls `calc_errors`.  In this way,  an application can directly use the point-generating function to implement a statistical error calculation, without using the fitplugin architecture.

### DIFF
--- a/bumps/errplot.py
+++ b/bumps/errplot.py
@@ -74,12 +74,9 @@ def reload_errors(model, store, nshown=50, random=True):
     return calc_errors_from_state(
         problem, state, nshown=nshown, random=random)
 
-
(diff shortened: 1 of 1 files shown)


bmaranville:
Note that this change is backward-compatible with existing applications that use a fitplugin and `calc_errors_from_state` - that function will still do the same thing it did before.

PR #184, opened 2024-11-13
adding helper script to build or remove stepper dll in deployed bumps

bmaranville:
For applications where bumps is installed (e.g. in refl1d, on a cluster), the manual process of compiling the stepper adds complexity to an otherwise simple `pip install bumps`.

In this PR, a builder script is added so that the optimized stepper can be built with:
```sh
python -m bumps.dream.build_compiled
```
with these optional flags:
```
$ python -m bumps.dream.build_compiled --help
usage: build_compiled.py [-h] [--nthreads NTHREADS] [--no-openmp] [--remove]

options:
  -h, --help           show this help message and exit
  --nthreads NTHREADS  Number of threads to compile for, use 0 for all available cores, default is 64
  --no-openmp          Disable OpenMP (e.g. for clang, where it is not supported)
  --remove             Remove the compiled dll
  --dry-run            Print the compile command without running it
```


bmaranville:
...got verbal approval for this change from @pkienzle... merging now.

PR #185, opened 2024-11-15
Various minor changes

glass-ships:
- [x] Rename `.mjs` files to `.ts` and adjust imports
    - This is because to use module files with typescript, you also need to have `.d.ts` files which declare your modules and add type hinting, otherwise you end up with silent errors/unexpected behavior. This change maintains functionality, but addresses some `vue-tsc` type testing errors in Refl1D
- [x] use `attribute!: Type` syntax on class `FileBrowserSettings` to indicate to TS that these attributes are required to be provided on instantiation, and thus no default value is needed
- [x] Adds vague `on()` and `off()` parameters to `AsyncSocket`, primarily to satisfy IDE linters (`emit()` _may_ need to be added as well.
- [x] `defineProps` no longer needs to be imported
- [x] Rename `plotcache` to `plot_cache` for consistency
- [x] Should address `asyncSocket.ts:5:9 - error TS2300: Duplicate identifier 'Socket'.` but I honestly don't understand how this chunk of code works yet. May need input/help from @bmaranville 
- [x] Should notify user if there's an issue in the draw loop


bmaranville:
...so the declaration in `asyncSocket.ts` is meant to add two methods to the existing `Socket` class defined by `socket-io`, since we don't directly instantiate that class, but rather use the imported `io` method from `socket-io`, which returns an instance of `Socket`.

I actually struggled with how to do this, and the code works even though typescript linters complain about it.  I'm sure there's a better way.

glass-ships:
I would not have thought switching to an interface would solve it. good find!

glass-ships:
Workflow wise, I'd recommend collecting merges into develop and then merging develop into master at various points, rather than merging directly into master and then having to rebase develop. 

To that end, we might want to consider bringing this in before too much else changes in master, provided it's still in a functional spot

bmaranville:
sure... squash merge?  Will we miss the commit messages?

bmaranville:
I was able to manually rebase on master with just two minor fixes... I might revert your merge and force-push that instead.

glass-ships:
probably not - I'd say we can squash. a lot of the merge commits and stuff will just be noise, and I believe auto-generating release notes still includes the highlights from each one (?)

bmaranville:
It looks like the merge is messed up in this branch as well - there are errors popping up for me after I pull the latest 'develop'

glass-ships:
Ah poop. What kinda errors?

bmaranville:
This is what I've had to fix so far:
```diff
diff --git a/bumps/webview/client/src/App.vue b/bumps/webview/client/src/App.vue
index 1bb275ce..415ae100 100644
--- a/bumps/webview/client/src/App.vue
+++ b/bumps/webview/client/src/App.vue
@@ -10,6 +10,7 @@ import {
   active_fit,
   selected_fitter,
   default_fitter,
+  default_fitter_settings,
   fitOptions,
   fileBrowser,
   FileBrowserSettings,
@@ -67,10 +68,6 @@ const socket = io(sio_server, {
 }) as AsyncSocket;
 socket_ref.value = socket;
 
-const can_mount_local = (
-  ('mountLocal' in socket) && ('showDirectoryPicker' in window)
-)
-
 socket.on('connect', async () => {
   console.log(socket.id);
   connected.value = true;
@@ -265,10 +262,6 @@ file_menu_items.value = [
   { text: "Quit", action: quit },
 ]
 
-if (can_mount_local) {
-  file_menu_items.value.splice(0, 0, { text: "Mount Local Folder", action: mountLocal });
-}
-
 </script>
 
 <template>
diff --git a/bumps/webview/client/src/app_state.ts b/bumps/webview/client/src/app_state.ts
index 24e48472..c0511a02 100644
--- a/bumps/webview/client/src/app_state.ts
+++ b/bumps/webview/client/src/app_state.ts
@@ -37,7 +37,7 @@ export const default_fitter_settings = shallowRef<{ [fit_name: string]: FitSetti
 export const selected_fitter = ref<string>();
 export const default_fitter = "amoeba";
 export const notifications = ref<{ title: string, content: string, id: string, spinner: boolean }[]>([]);
-export const file_menu_items = ref<{disabled?: Ref<boolean> | boolean, text: string, action?: () => void, help?: string}[]>([]);
+export const file_menu_items = ref<{disabled?: Ref<ComputedRef<boolean>> | boolean, text: string, action?: Function, help?: string}[]>([]);
 export const autosave_history = ref(false);
 export const autosave_history_length = ref(10);
 export const session_output_file = shallowRef<{ filename: string, pathlist: string[] }>();
```

bmaranville:
I think these are all carryovers from the previous PR merge?

glass-ships:
looks like it yeah

bmaranville:
This is making me a little nervous... all sorts of merge errors that aren't immediately noticeable could slip through.  I only found these because the linter saw them.

glass-ships:
yeah definitely part of what i was worried about :/ 
+1 for linting though for sure, definitely motivates getting that pulled in sooner rather than later too

glass-ships:
Also i think 
```
export const file_menu_items = ref<{disabled?: Ref<ComputedRef<boolean>> | boolean, text: string, action?: Function, help?: string}[]>([]);
```
should be

```
export const file_menu_items = ref<{disabled?: ComputedRef<boolean> | boolean, text: string, action?: Function, help?: string}[]>([]);
```
no?

bmaranville:
> Also i think
> 
> ```
> export const file_menu_items = ref<{disabled?: Ref<ComputedRef<boolean>> | boolean, text: string, action?: Function, help?: string}[]>([]);
> ```
> 
> should be
> 
> ```
> export const file_menu_items = ref<{disabled?: ComputedRef<boolean> | boolean, text: string, action?: Function, help?: string}[]>([]);
> ```
> 
> no?

No, Vue (or TypeScript?) does something slightly funny with Ref objects inside other Ref objects - they are automatically unwrapped in some contexts.  Try it without the additional Ref<> wrapper - you'll see an error in App.vue when we set  `disabled` to a computed ref.

glass-ships:
hmmmm. Locally, I see this when the Ref<> wrapper is included, but removing it makes this error go away
image

bmaranville:
I think yours might be doing a better check than mine, though, because mine is showing `type: any` for the disabled attribute in the template.

glass-ships:
woah, weird..

glass-ships:
let's compare vs code extensions tomorrow

PR #188, opened 2024-11-18
Remove bootstrap js

bmaranville:
This PR removes the Javascript parts of Bootstrap from the webview client.
The client still uses primarily Bootstrap styles (through the CSS definitions) for display, so the bootstrap package dependency remains.

From the beginning of the development of the client, it has been awkward to mix using Vue.js and Bootstrap.js when rendering dynamic components.  The remaining items that we were still using Bootstrap.js for:

- dropdown menus, with the `data-bs-toggle="dropdown"` attribute (the presence of this attribute signals the Bootstrap runtime to add event handlers to this element)
- similarly, the behavior of the main menu (collapsed or not collapsed on small screens) was controlled by a slightly different `data-bs-toggle="collapsed"` attribute
- all of our modal dialogs were using an instance of the `Modal` object from Bootstrap.js

After this PR

- dropdown menus are handled through a newly-written `DropDown.vue` component
- the main menu collapsed value is handled by a simple Vue callback, toggling a boolean Ref
- modal dialogs are handled natively by the newly standardized HTML5 <dialog> element (still styled with Bootstrap CSS)


glass-ships:
I'm getting some issues with this locally after resolving merge conflicts

image

bumps/webview/client/src/components/FitOptions.vue
```diff
@@ -1,7 +1,6 @@
 <script setup lang="ts">
-import { ref, onMounted, computed, shallowRef } from 'vue';
-import { fitter_settings, selected_fitter } from '../app_state.ts';
-import type { FitSetting } from '../app_state.ts';
+import { ref, onMounted, computed } from 'vue';
+import { default_fitter, default_fitter_settings, fitter_settings, selected_fitter } from '../app_state.ts';
```
bmaranville, line 3: These changes were in #186 that was merged 3 days ago... was the master you merged in not up to date?

glass-ships: i think master was, but i must have missed it in resolving the merge conflicts (since those changes weren't here yet, but i assumed whatever was in this branch was what we wanted)

bmaranville: For any merge that gets pushed in that requires manual resolution, it would be nice to have a comment added to the PR "this was manually merged".  Automatic merges don't usually need a detailed review but manual ones do.

glass-ships:
thanks for cleaning that up!

bmaranville:
I did some manual testing that the client is doing the right things (the different modals are popping up, and the dropdown menus are functioning correctly), so I think this is ready to go.

The merge on FitOptions.vue was always going to be hard, since the same code was modified extensively in two different PRs... I could have waited and based this one on the changes in the other, and that would have saved some trouble.

glass-ships (approved):

PR #190, opened 2024-11-19
Add Ruff, pre-commit, and a Makefile

glass-ships:


bumps/dream/diffev.py
```diff
@@ -39,7 +39,13 @@ def njit(*args, **kw):
 EPS = 1e-6
 _DE, _SNOOKER, _DIRECT = 0, 1, 2
 
+<<<<<<< HEAD
```
bmaranville, line 42: I'm surprised git let this get pushed... usually it stops me if a merge is not complete.

glass-ships: yeah i've been having some weird behavior with git the last few days... not sure what's up

glass-ships:
.... should `dream/Random123/doc` and stuff have made it into the repo here? 
that feels. like that should not have happened

bumps/dream/Random123/docs/cbrng.dox
```diff
@@ -24,7 +24,7 @@ error or substitute an alternative compatible
 CBRNG.)
 
 The API descriptions below are generic, but apply to
-all the different @ref families "families" of 
+all the different @ref families "families" of
```
bmaranville, line 27: no formatting should be applied to any file in the Random123 folder.  It is a vendored dependency and not part of our source management.

glass-ships:
Closing in favor of #193

PR #191, opened 2024-11-20
Add linting and formatting for webview client

glass-ships:
- Adds prettier and eslint w/ scripts for webview client code
- Adds github action to test lint
- Adds precommit for eslint but ignore hook in ci


glass-ships:
Git UI is saying there are conflicts for multiple files, but running `git merge master` locally just returns "Already up to date." so I'm not sure what the deal there is

glass-ships:
It looks like `error_text` is only being generated by the client and displayed:
```typescript
async function fetch_and_draw() {
  ...
  } else if (fig_type === "error") {
    error_text.value = String(plotdata).replace(/[\n]+/g, "<br>");
  } else {
    figtype.value = "error";
    error_text.value = `Unknown figure type: ${fig_type}`;
  }
}
```

Whether this should still be sanitized I'm not too sure (I don't understand XSS mitigation that well) 

[This cheatsheet](https://cheatsheetseries.owasp.org/cheatsheets/Cross_Site_Scripting_Prevention_Cheat_Sheet.html#html-sanitization) recommends [DOMPurify](https://github.com/cure53/DOMPurify) for HTML sanitization, but it may or may not really be necessary in our case.

bumps/webview/client/src/components/ModelInspect.vue
```diff
@@ -1,63 +1,56 @@
...
+  } else {
     // do update of existing model:
     const old_model: json = modelJson.value as Record<string, any>;
     const new_model: json = json_value as Record<string, any>;
     const diff = getDiff(old_model, new_model);
     for (let [path, oldval, newval] of diff.edited) {
       const { target, parent, key } = resolve_diffpath(old_model, path);
       // trigger reactive update;
-      if (typeof (key) === 'number' && Array.isArray(parent)) {
+      if (typeof key === "number" && Array.isArray(parent)) {
```
bmaranville, line 40: I feel like someday JS is going to do the same thing python did with print -> print() on typeof and instanceof... but today is not that day (and the linter will save us that day!)

.github/dependabot.yaml
```diff
@@ -0,0 +1,9 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      # Check for updates to GitHub Actions every week
+      interval: "weekly"
```
bmaranville, line 9: They don't change that fast... would prefer monthly

.pre-commit-config.yaml
```diff
@@ -0,0 +1,38 @@
...
+      - id: eslint
+        name: eslint
+        entry: bash -c "cd bumps/webview/client && npm run lint"
+        language: system
+  - repo: https://github.com/pre-commit/pre-commit-hooks
+    rev: v4.6.0
+    hooks:
+      - id: check-added-large-files
+        args: [--maxkb=8192]
+        exclude: bumps/dream/Random123
```
bmaranville, line 13: I think everywhere it says "Random123" right now it should be "random123", since the PR that made it an external dependency

bmaranville (approved):
Looks good - I think the only critical change needed is to the "Random123" ignore rules.

.pre-commit-config.yaml
glass-ships: image

any of these instances as well? or just the precommit config

bmaranville: wow - that's a good catch.  I am confused - I changed almost all those instances as part of the PR that moved random123 to an external dependency, and I don't see where those changes were reverted, but the current master doesn't match the PR files that were merged... e.g. https://github.com/bumps/bumps/pull/195/files#diff-41d5a52589e0480be9c099d2bba7a8135b8b0d71bcbb8df3582a8df1c2295003

bmaranville: Except the first one - the library is in fact named `Random123` even though the repo is named `random123` (and therefore that is the name of the outer directory)

glass-ships: cool, just pushed a commit - let me know if anything looks not quite right

PR #193, opened 2024-11-22
Add Ruff, pre-commit, and a Makefile

glass-ships:
Includes:
- automated formatting with ruff
- manual option to lint with ruff
- Makefile with various targets
- pre-commit config for automated formatting on commits

New version of #190 to avoid possible mismatches from various merges


bmaranville:
I think we'll have to wait on this one until after 1.0 sometime - we'll have to be careful to squash it and hide it from git blame like we discussed.

.pre-commit-config.yaml
```diff
@@ -14,24 +14,24 @@ repos:
...
+  - repo: https://github.com/astral-sh/ruff-pre-commit
+    rev: v0.6.1
+    hooks:
+      # - id: ruff
+      #   args: [--fix, --exit-non-zero-on-fix]
+      - id: ruff-format
+        args: [--config=pyproject.toml]
+        exclude: |
+          bumps/dream/random123
+          bumps/webview/client/src
```
bmaranville, line 34: We need to exclude everything in extra/platform_scripts as well - and all those files have to be reverted (the line-endings are important)

extra/sasview/FitPage2.fitv
```diff
@@ -291,12 +291,12 @@
 		</SASdata>
 		<SASsample name="">
 			<ID>
-				
+
```
bmaranville, line 294: I don't really know if this file is still used or useable, but we shouldn't be reformatting a serialized output file like this.  It should be reverted.

bmaranville (approved):
I withdraw my previous objection... the sasview example is not working at all (the imports from sasview have all changed), so bringing that up to date would be the subject of a completely different PR.  Let's not worry about it here.

PR #200, opened 2024-12-06
Remove client cdn rebase

bmaranville:
replaces #172, rebased on current master.


glass-ships:
so this requires running the script which builds bumps within the refl1d space? (where does bumps get build to relative to refl1d web view code

bmaranville:
> so this requires running the script which builds bumps within the refl1d space? (where does bumps get build to relative to refl1d web view code?)

The only thing this changes is that it builds the client in `dist/` directly instead of `dist/<version-string>/`

It doesn't alter the client source at all, and it simplifies the python code in `webserver.py` because finding the client code to serve is a lot easier.

glass-ships:
ahhh I see ok!

glass-ships (approved):

bmaranville:
There's a similar PR in refl1d that I will now look at, because it should be merged right after this one.

PR #201, opened 2024-12-06
Fix gaussian mixture example

pkienzle:
The gaussian mixture example demonstrates that the dream fitter can return the correct posterior distribution.

It was broken in python 3 since zip returns an iterator that is being looped multiple times. The iterator is now turned into a tuple.

A slight tweak to the plotting for when there are bad values for log probability: clear these values before creating the histogram.

### DIFF
--- a/bumps/dream/model.py
+++ b/bumps/dream/model.py
@@ -242,4 +242,5 @@ class Mixture(MCMCModel):
 
     def nllf(self, x):
         p = [w*exp(-M.nllf(x)) for M, w in self.pairs]
-        return -log(np.sum(p)/self.weight)
+        total = np.sum(p)/self.weight
+        return -log(total) if total > 0 else 1e100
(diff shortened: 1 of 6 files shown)


bumps/dream/model.py
```diff
@@ -237,9 +237,10 @@ def __init__(self, *args):
...
             raise TypeError("Expected MixtureModel(M1, w1, M2, w2, ...)")
-        self.pairs = zip(models, weights)
+        self.pairs = tuple(zip(models, weights))
         self.weight = np.sum(w for w in weights)
 
     def nllf(self, x):
         p = [w*exp(-M.nllf(x)) for M, w in self.pairs]
-        return -log(np.sum(p)/self.weight)
+        total = np.sum(p)/self.weight
+        return -log(total) if total > 0 else 1e100
```
bmaranville, line 246: That's a big log number.

bmaranville (approved):

PR #202, opened 2024-12-09
Webview Client - add type testing with vue-tsc

glass-ships:
Adds type testing with `vue-tsc`


bumps/webview/server/api.py
```diff
@@ -589,95 +765,188 @@ async def get_custom_plot_info():
...
-            fig['data'].append(dict(type="scattergl", x=x, y=pop[tail:,2].tolist(), name="population median", mode="lines", line=dict(color="green"), opacity=0.5))
-            fig['data'].append(dict(type="scattergl", x=x, y=pop[tail:,0].tolist(), name="population best", mode="lines", line=dict(color="green", dash="dot")))
-
-        fig['data'].append(dict(type="scattergl", x=x, y=best[tail:].tolist(), name="best", line=dict(color="red", width=1), mode="lines"))
-        fig['layout'].update(dict(template="simple_white", legend=dict(x=1,y=1,xanchor="right",yanchor="top")))
-        fig['layout'].update(dict(title=dict(text="Convergence", xanchor="center", x=0.5)))
-        fig['layout'].update(dict(uirevision=1))
-        fig['layout'].update(dict(xaxis=dict(title="iteration number", showline=True, showgrid=False, zeroline=False)))
-        fig['layout'].update(dict(yaxis=dict(title="chisq", showline=True, showgrid=False, zeroline=False)))
+            fig["data"].append(
```
bmaranville, line 848: I find this a lot harder to read than the original... way too many lines.  It seems like the formatter for typescript collapsed a lot of things to one line where the formatter for python expanded all the one-liners - why is that?

glass-ships: ruff and prettier do have different conventions on occasion, while ruff is a lot more configurable. 

in this case what it comes down to though is that a lot of these lines are over the 120 char line limit. 

in the cases that are shorter, i _believe_ the arguments to the `dict()` method can be forced to one line by omitting the trailing comma. see line 910 as an example. 

i agree it can seem a little bloated at first, especially when trying to compare multiple lines to each other, but it does also make analyzing the arguments for one particular function a bit easier

bmaranville:
Some of the API calls would have thrown errors when trying to call `syncFS`, which was declared as a method on the interface, but wasn't implemented anywhere.

glass-ships:
i was wondering what to do about that!

bumps/webview/client/src/components/LogView.vue
```diff
@@ -8,13 +8,21 @@ const props = defineProps<{
...
 
 props.socket.on("log", ({ message: { message, title } }) => {
   log_info.value.push({ message, title });
 });
-props.socket.asyncEmit("get_topic_messages", "log", (messages) => {
-  log_info.value = [...log_info.value, ...messages.map((m) => m.message)];
+
+props.socket.asyncEmit("get_topic_messages", "log", (messages: TopicMessages) => {
+  console.log({ messages });
+  log_info.value = [...log_info.value, ...messages.map((m) => ({ message: m.message, title: m.title }))];
```
bmaranville, line 25: This isn't quite right... log messages are not showing up correctly on connect.

glass-ships: how are they showing up? 

could this be a result of how the python api is returning the message object from `get_topic_messages`? i took a look at the backend code, but that method was pretty confusing

bmaranville: They are showing up as text representations of the object with keys above, e.g. 
```
{ "message": "Loading model: /home/bbm/dev/refl1d-modelbuilder/ISIS_GGG/GGG_GdIG_MultiFit_bbm.py", "title": null }
{ "message": "Model loaded: /home/bbm/dev/refl1d-modelbuilder/ISIS_GGG/GGG_GdIG_MultiFit_bbm.py", "title": null }
```

The change suggested in the PR is a logic change, not a formatting or type change - the map is returning 
```ts
messages.map((m) => ({ message: m.message, title: m.title }))
```
instead of 
```ts
messages.map((m) => m.message
```

glass-ships: As an update - I looked through the api code.  
There is a defined `log()` method which adds a message to the "topics" state attribute.  
But I don't see anywhere that calls `log()` on connect/disconnect - which may be why we're not seeing anything in the Log tab

bmaranville: The log panel retrieves any stored log messages when it opens: 
https://github.com/bumps/bumps/blob/07d1b185fccf11dee6dc45b0d90cbd8e8e08ea57/bumps/webview/client/src/components/LogView.vue#L16

bumps/webview/client/src/components/UncertaintyView.vue
```diff
@@ -47,19 +49,19 @@ async function fetch_and_draw(latest_timestamp: string): Promise<void> {
...
   layout.showlegend = visible;
   for (let item_name in layout) {
     if (item_name.startsWith("xaxis")) {
-      layout[item_name].showticklabels = visible;
+      (layout as any)[item_name].showticklabels = visible;
     }
   }
   for (let annotation of layout?.annotations ?? []) {
-    if (annotation?.name === "label") {
+    if (annotation?.text === "label") {
```
bmaranville, line 64: This change seems incorrect - why change `name` to `text`?  The contents of the `annotation` object coming from the server haven't changed.

bmaranville: Ah - it looks like I added a `name` property to the annotation object on the Python side, for bookkeeping - to distinguish the labels from the other annotations.  `name` is not a part of the defined interface for `Plotly.Annotations`.  But again here, the logic is correct and the types are wrong.  We shouldn't change the logic any time a type conflict arises unless we verify there is a logical bug.

bumps/webview/server/api.py
bmaranville: I agree - overall the formatting makes the code easier to read.  There are times that the line limit seems unhelpful, but I suppose I can always add pragma statements to ignore those lines if I think it's important.

bumps/webview/client/src/components/UncertaintyView.vue
glass-ships: ah i see, i didn't realize we had another case of extending an existing interface - i'll go ahead and fix that

bumps/webview/server/api.py
```diff
@@ -758,24 +1050,34 @@ async def get_parameter_labels():
...
         start_time = time.time()
-        logger.info(f'queueing new parameter_trace plot... {start_time}')
+        logger.info(f"queueing new parameter_trace plot... {start_time}")
 
         # begin plotting:
         portion = None
         draw, points, _ = uncertainty_state.chains()
         label = uncertainty_state.labels[var]
-        start = int((1-portion)*len(draw)) if portion else 0
-        genid = np.arange(uncertainty_state.generation-len(draw)+start, uncertainty_state.generation)+1
```
bmaranville, line 776: Ok, this one I don't understand:  the line length is well under 120, and this line makes far more sense to read as a one-liner than in multiple lines with a ` + 1 ` on a line by itself.  Which rule is triggering this?

bmaranville (approved):

bumps/webview/server/api.py
glass-ships: huh... good point, you know i'm not sure! formatters definitely have their interesting preferences on splitting functions and statement segments into new lines vs collapsing them

PR #203, opened 2024-12-09
Update vite.config.js

glass-ships:
It looks like plotly (indirectly, via `has-hover`) requires a definition of `global` to work, otherwise the page won't load


bmaranville (approved):
Ah... I chopped too much.

PR #204, opened 2024-12-12
Add application schema to serialized metadata, helps with application model migration

bmaranville:
This PR 
- adds a new `libraries` property to the serialized JSON format for models, which provides a `version` string and `schema` string for the libraries from which serialized dataclasses are made for a particular model.
- adds a new `migrate_serialized(serialized_obj: bytes)` method to the interface of `bumps.plugin`, to support application-specific model migration prior to deserialization.

During the serialization, a library of package metadata is created.  As the model tree is walked, as each dataclass encountered is inspected to see which python package corresponds to the module and class it represents.  If that package has not yet bet added to the metadata, the version is extracted from the package and the `__schema_version__` attribute is read from the base module of the class path, to populate an entry 
```
"libraries": { 
    "package-name": { 
        "version": "version_string", 
        "schema": "schema_string or None",
    }
}
```

This dictionary is also written to the HDF serialization of the webview session file, in the "problem" group (readable without deserializing the problem)

### DIFF
--- a/bumps/plugin.py
+++ b/bumps/plugin.py
@@ -38,6 +38,7 @@ __all__ = [
     'show_errors',
     'data_view',
     'model_view',
+    'migrate_serialized',
 ]
 
 # TODO: refl1d wants to do the following after cli.getopts()
@@ -110,3 +111,13 @@ def model_view():
     Return None if not present.
     """
     return None
+
+
+def migrate_serialized(model_dict):
+    """
+    Migrate serialized model to the current version.
+
+    This function is called when loading a model from a file.  It is
+    used to update the model to the current version of the plugin.
+    """
+    return model_dict
--- a/bumps/__init__.py
+++ b/bumps/__init__.py
@@ -17,6 +17,8 @@ try:
 except ImportError:
     __version__ = "unknown"
 
+__schema_version__ = "1"
+
 def data_files():
     """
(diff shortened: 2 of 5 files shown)


bumps/plugin.py
```diff
@@ -110,3 +111,13 @@ def model_view():
...
+
+
+def migrate_serialized(model_dict):
+    """
+    Migrate serialized model to the current version.
+
+    This function is called when loading a model from a file.  It is
+    used to update the model to the current version of the plugin.
+    """
+    return model_dict
```
glass-ships, line 123: So this is an "abstract" function meant to do nothing unless overloaded by the plugin, correct?

glass-ships (commented):
were you able to test locally that this works alongside the refl1d PR?

bumps/plugin.py
bmaranville: Correct

bmaranville:
> were you able to test locally that this works alongside the refl1d PR?

Yes, and I was able to load an old stored session (even one that needed both the bumps migration and the refl1d migration)

glass-ships (approved):

glass-ships:
> Yes, and I was able to load an old stored session (even one that needed both the bumps migration and the refl1d migration)

cool! looks good to me

PR #211, opened 2025-01-27
Remove deprecated div

bmaranville:
Some old models have serialized an expression with a division operation as `op: 'div'`, which means that they had the `__div__` method in their class (which means they were using what is now an old Python version, as `__div__` was deprecated a while back).

New models will have `op: 'truediv'` for floats or `op: 'floordiv'` for integer division.

This PR adds a migration to the bumps deserialization to convert instances of `op: 'div'` to `op: 'truediv'`


bumps/serialize.py
```diff
@@ -23,8 +23,9 @@
 class SCHEMA_VERSIONS(str, Enum):
```
pkienzle, line 23: How does `SCHEMA_VERSIONS(str, Enum)` differ from `SCHEMA_VERSIONS(StrEnum)`?

pkienzle (approved):

bumps/serialize.py
glass-ships: `StrEnum` is just a newer implementation in python 3.11, the end result is the same. 
Doing it as `(str, Enum)` allows it to be compatible with older versions of python

PR #217, opened 2025-02-20
Generic fitproblem

bmaranville:
Make FitProblem a Generic class, so that a consuming application can specify the type of models.

For example: in Refl1D, we could do the following:
```python
from bumps.fitproblem import FitProblem
from refl1d.experiment import Experiment

Refl1DFitProblem = FitProblem[Experiment]
```
and then the objects returned from the generator `Refl1DFitProblem.models` would be expected to be of type `Experiment`

### DIFF
--- a/bumps/fitproblem.py
+++ b/bumps/fitproblem.py
@@ -51,6 +51,7 @@ import logging
 import os
 import sys
 import traceback
+from typing import Generic, TypeVar
 import warnings
 
 import numpy as np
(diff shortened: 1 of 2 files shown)


glass-ships:
Will there need to be a corresponding Refl1d PR making use of this?

bmaranville:
It is not required - the typing parameter is optional of course (it is the Fitness protocol by default).  But we can take advantage of it, and that's what's in the refl1d PR 258 mentioned above.

glass-ships:
Ah nice ok! 
Only other question - I've never understood the `Generic` class.. can you explain what `Generic[FitnessType]` is doing as far as typing?

glass-ships (approved):

bmaranville:
In newer Python (>=3.12?) you can write this down a lot more easily, without the `Generic` class - e.g.
```python
class FitProblem[FitnessType]:
    models: List[FitnessType]
    ...
```
Which lets you supply a type parameter to FitProblem that describes what will go into the models attribute.  Then in consumer code where e.g. `problem = FitProblem[Experiment](...)`, your type checker will know type `model` is below

```python
for model in problem:
    # model is of type Experiment
```

PR #218, opened 2025-03-03
Shared memory for populating MPMapper workers

bmaranville:
Currently a synchronized `Namespace` from a `multiprocessing.Manager` object is used to transmit the fit problem to the workers in an MPMapper pool.  It follows two different paths:

- if `dill` is available, it is used to serialized the problem to a `bytes` object which is stored in the `Namespace`
  - then in the workers, if `dill` is available, it is used to deserialize
- else, the problem is stored as-is in the `Namespace`

In both cases, `pickle.dumps` is called on the contents of the `Namespace` before sending to the workers (which is nearly a no-op for a `bytes` object) and `pickle.loads` is called on the receiving event, automatically, by the manager/`Namespace` infrastructure.

In this PR, the `Namespace` object is removed, and a shared-memory `manager.Array("B"...)` is used to share a serialized problem with the workers as a `bytes` object.   When `dill` is available, it is used to serialize, otherwise `pickle` is used **explicitly** (rather than implicitly by a `Namespace`).

The loading of a new problem object is triggered the same way as before, by incrementing the `problem_id` that is sent in each evaluation call.


bmaranville:
It's a pretty minor optimization unless your serialized fitproblem gets big, but I found it made a difference when parallelizing the plots.

pkienzle (approved):
I see no problems with the code. Assuming our CI tests this on Mac/Windows/Linux (does it?) then apply as is. All changes are internal. I don't know the relative merits of Namespace and SharedMemory. Speed is immaterial since problem transfer is rare. So go with what is most reliable and maintainable.

Do we want to simplify the code by requiring dill?

PR #221, opened 2025-03-07
Binary installers

bmaranville:
Change infrastructure to build binary installers for MacOS and Windows rather than distributing compressed archives of the packed folders directly.

For linux, the compacted folder still seems to be a reasonable distribution method (tar is available, digital signatures not typically required)

This PR results in building 
- a MacOS dmg signed with our developer certificate, which contains:
  - a copy of the template in `extra/platform_resources/macos/app_folder_template`
    - bumps.app
    - bumps_shell.app
    - a shortcut to the `/Applications` folder, for ease of install
  - with the contents of the conda-packed bumps environment copied to `bumps.app/Contents/Frameworks/python.framework/Resources/env`
- a Windows installer (made with NSIS)
  - also with the contents of the conda-packed environment, as the src of installed files
  - a shortcut which starts the webview server in a Powershell session
  - a shortcut which starts a Powsershell interactive session with the bumps environment activated (with micromamba), so that users can install more packages if they want.

Note that the Windows installer is not signed (we don't have a Windows certificate yet)

On all platforms, the "bumps-webview" command works as expected (in Windows, the conda-unpack command is run during installation, while on linux and macos the bumps-webview executable script already does the right thing)


extra/installer.nsi
```diff
@@ -0,0 +1,108 @@
...
+!include "WinVer.nsh"
+!include "x64.nsh"
+
+;-------------------------------------------------------------------------------
+; Constants
+!ifndef PRODUCT_NAME
+    !define PRODUCT_NAME "Bumps"
+!endif
+!define PRODUCT_DESCRIPTION "Bayesian Uncertainty Modeling for the Physical Sciences"
+!define COPYRIGHT "Copyright © 2018 The bumps developers"
```
pkienzle, line 14: Copyright 2011-2025 (under this name; you can go back to 2006 for the start of DANSE, or the mid 90's for some of the borrowed bits). The combined work is in the public domain, so "The Bumps developers"* are free to slap a 2024-2025 copyright on it.

\* Note the capitalization

extra/installer.nsi
```diff
@@ -0,0 +1,108 @@
...
+!define PRODUCT_VERSION 1.0.0.1013
+!define SETUP_VERSION 1.0.0.0
```
pkienzle, line 16: Version number hard-coded that needs to be changed for each point release?

extra/installer.nsi
```diff
@@ -0,0 +1,108 @@
...
+!ifndef SRC
+    !define SRC "..\conda_packed"
+!endif
+
+;-------------------------------------------------------------------------------
+; Attributes
+Name "${PRODUCT_NAME}"
+OutFile "BumpsWebviewSetup.exe"
+InstallDir "$LocalAppData\${PRODUCT_NAME}"
+InstallDirRegKey HKCU "Software\Reflectometry-Org\${PRODUCT_NAME}" ""
```
pkienzle, line 26: Reflectometry-Org? Or the International Scattering Alliance umbrella? There are more below in this file and in the mac plist.

.github/workflows/build-distributables.yml
```diff
@@ -31,28 +35,23 @@ jobs:
...
       env:
           MACOS_CERTIFICATE: ${{ secrets.MACOS_CERTIFICATE_ISA }}
           MACOS_CERTIFICATE_PWD: ${{ secrets.MACOS_CERTIFICATE_ISA_PWD }}
           KEYCHAIN_PASSWORD: ${{ secrets.KEYCHAIN_PASSWORD }}
           NOTARIZATION_USERNAME: ${{ secrets.MACOS_NOTARIZATION_USERNAME }}
           NOTARIZATION_PASSWORD: ${{ secrets.MACOS_NOTARIZATION_PASSWORD }}
+          DEVELOPER_IDENTITY: "Developer ID Application: The International Scattering Alliance (8CX8K63BQM)"
       run: |
         echo $MACOS_CERTIFICATE | base64 --decode > certificate.p12
         security create-keychain -p $KEYCHAIN_PASSWORD build.keychain
```
pkienzle, line 57: Github docs [here](https://docs.github.com/en/actions/security-for-github-actions/security-guides/using-secrets-in-github-actions#using-secrets-in-a-workflow) recommend against using `$SECRET` on the command line.

The security command does have an interactive mode so it may be possible to feed in the secrets by echoing into stdin.

I don't understand the attack surface. Anyone who can modify and trigger this build script can get the secrets without too much trouble. Maybe if we are running some third party action which manages to persist throughout the workflow, but they don't need the secrets since they can inject a trojan directly into our package before we sign it.

I say ignore this for now, but we should at least wrap "$SECRET" in double quotes in case there are shell characters inside.



I'm not sure

pkienzle (approved):
Minor changes requested.

No idea how you tested this infrastructure since it is triggered for master on a published release.

extra/installer.nsi
bmaranville: We can put `!ifndef...!endif` around the version too, if you want, then you can pass it in as a variable at build time with `/DPRODUCT_VERSION=...`

extra/installer.nsi
bmaranville: Current guidance seems to suggest not putting an end date for copyright

PR #225, opened 2025-03-20
Dark mode

bmaranville:
Add a button to enable dark mode for the webview client.  This capability is built-in to the Bootstrap CSS we are using, we just have to expose it by setting e.g. `document.documentElement.setAttribute("data-bs-theme", "dark")`

Result:
image

### DIFF
--- a/bumps/webview/client/src/App.vue
+++ b/bumps/webview/client/src/App.vue
@@ -39,10 +39,15 @@ const props = defineProps<{
 }>();
 
 const show_menu = ref(false);
-const dark_mode = window.matchMedia("(prefers-color-scheme: dark)");
-if (dark_mode.matches) {
-  document.documentElement.setAttribute("data-bs-theme", "dark");
+
+// Control Dark Mode from system media query
+const prefersDarkMediaQueryList = window.matchMedia("(prefers-color-scheme: dark)");
+function setDarkTheme(prefersDark: boolean) {
+  const theme = prefersDark ? "dark" : "light";
+  document.documentElement.setAttribute("data-bs-theme", theme);
 }
+prefersDarkMediaQueryList.addEventListener("change", (e) => setDarkTheme(e.matches));
+setDarkTheme(prefersDarkMediaQueryList.matches);
 
 // const nativefs = ref(false);
 
--- a/bumps/webview/client/src/App.vue
+++ b/bumps/webview/client/src/App.vue
@@ -39,7 +39,11 @@ const props = defineProps<{
 }>();
 
 const show_menu = ref(false);
-const dark_mode = ref(false);
+const dark_mode = window.matchMedia("(prefers-color-scheme: dark)");
+if (dark_mode.matches) {
+  document.documentElement.setAttribute("data-bs-theme", "dark");
+}
+
 // const nativefs = ref(false);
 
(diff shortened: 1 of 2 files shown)


pkienzle (commented):
Instead of a setting, can we just ask the browser?

```javascript
const useDarkMode = window.matchMedia('(prefers-color-scheme: dark)');
```
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

The white background in the plots are jarring in dark mode. Replacing white with transparent or black for the background will lead to contrast problems for the lines and symbols. Can we set the background to white with  0.8 alpha? This will become light gray in dark mode, so not quite so jarring, but still light enough that the plot elements will show up.

I would rather not clutter the interface with a dark mode setting. If we have the setting, then we need storage or cookies to preserve it. Then we can hide it in a menu since we won't need to toggle it every time we touch the application.

bmaranville:
The feature was requested by a user - they didn't mention anything about wanting it to be persistent as a setting.  Since it's easy to enable I'd rather provide some path for them to do it, though we can hide it if you don't want to clutter things up.

bmaranville:
I can't quickly figure out how to turn on dark mode at the browser level in Chromium so I'm guessing this isn't a trivial thing in general in other browsers either

pkienzle:
Settings > search dark > theme : light, dark, device

I tested the above code with dark mode set and it returned the right thing.

I figure anyone who wants dark mode knows how to set it in their browser already.

Next question is what happens when dark mode is set to device?

pkienzle:
When preference is device then the above code matches light for me, and doesn't match device, so that's all you need.

For completeness may want to detect scheme changes [untested]:
```javascript
function updateTheme(event) {
    const isDark = event.matches;
    ...
}
window.matchMedia('(prefers-color-scheme: dark)').addEventListener('change', updateTheme);
```

pkienzle:
On MacOS  Settings > Appearance > Light | Dark | Auto

bmaranville:
This works on Firefox in linux by setting the browser dark mode, but in Chrome/Chromium you have to set the OS mode to dark apparently.

bmaranville:
I'm not sure the mpld3 plots show up well in dark mode... the axes and labels are black on dark grey.  Might want to figure out some extra styling for the data view.

pkienzle:
I tested the following on firefox and it responds to toggling the theme for the gnome control panel:
```javascript
window.matchMedia('(prefers-color-scheme: dark').addEventListener('change', e => console.log("dark", e.matches));
```
Windows and Mac also support light/dark themes.

bmaranville:
image

bmaranville:
I don't think we really need it to switch in real time.  If they want they can reload the page.

pkienzle:
Interesting... with CSS it automatically follows the light/dark selection in the MacOS settings. Can you do the other changes with CSS as well?

pkienzle:
Some people have it set up to automatically switch on a schedule or depending on ambient light with mobile devices.

Unfortunately if it starts in dark mode and switches to light mode, the axes become unreadable.

image

bmaranville:
no, the other change is to an attribute of the documentElement (not CSS)

bmaranville:
Ok - it tracks dynamically now

pkienzle (approved):

PR #227, opened 2025-03-21
scroll behavior on sliders is more harmful than helpful

bmaranville:


bmaranville:
@pkienzle mentioned that the default scroll behavior on the sliders was interfering with scrolling the content of the tab (seeing all the parameters in the summary tab) and I've noticed the same thing.

This PR prevents scrolling on the sliders from changing their value.  Once the slider is focused in the browser, keyboard arrows are still able to alter the value, so I think accessibility is not affected.

pkienzle (approved):

PR #229, opened 2025-03-21
Default install client dependencies

bmaranville:
add --no-deps flag (by default, False) to bumps.webview.build_client

This change means that `npm install` is run by default when building the client with the python helper, rather than not being run by default.

If no `node_modules` folder is found, `npm install` is still run no matter what the --no-deps flag value is.

### DIFF
--- a/bumps/webview/build_client.py
+++ b/bumps/webview/build_client.py
@@ -4,7 +4,7 @@ import shutil
 
 
 def build_client(
-    install_dependencies=False,
+    no_deps=False,
     sourcemap=False,
     cleanup=False,
 ):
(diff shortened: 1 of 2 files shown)


pkienzle (approved):
Tested and works.

Seems to bundle the purge commits. It'll probably apply cleanly even though I've already merged them to master.

pkienzle:
Note: need to add client/dist and node_modules to .gitignore

bmaranville:
I didn't base it cleanly at first... I just rebased and force pushed and it's better now.

bmaranville:
> Note: need to add client/dist and node_modules to .gitignore

node_modules is already in .gitignore, but you're right about client/dist

pkienzle:
... and package-lock.json

bmaranville:
This is cleanup that we did for refl1d but never did for bumps.

pkienzle:
Ahhh... hiding at the end. I wondered why it wasn't showing up in git status.

PR #231, opened 2025-03-26
Autohide panels

bmaranville:
Create new shared state attributes that indicate if, e.g. these plots are available

- uncertainty plots
- custom plots
- custom uncertainty plots

Also refactors shared state subscriptions in the client with a single shared state object used by components, that receives updates from the server.


pkienzle:
Blank model uncertainty not hidden in dream fit. Try with:
```sh
python -m bumps.webview.server.webserver doc/examples/curvefit/curve.py --fit=dream --start
```

bmaranville:
It only hides panels when uncertainty is not available... It doesn't check to see if model uncertainty is implemented.  We can probably unify the model uncertainty with custom plots

pkienzle (approved):

PR #236, opened 2025-04-01
Disconnected dialog refactor

bmaranville:
Currently, if the user quits the server (or it is terminated) any client window that is connected to that server shows a dialog with "Server shutting down" and a "DISMISS" button.

The button is supposed to prevent the tab/window from auto-closing, but this is not obvious to the user based on the dialog and button wording.  

Also, currently if the user clicks "DISMISS" the next thing they see is another dialog with "Server connecting..." and a spinner that spins endlessly, but even if another server is started at the same hostname/port, the client never reconnects.

This PR fixes the second issue (the client does attempt to reconnect), and also shows a more informative dialog when the server shuts down.

The PR removes the "Connected" and "Disconnected" indicators in the upper-right of the header, saving screen real estate.  A disconnected client shows a dialog indicating "Disconnected", preventing any interaction with the client parts, until a new server is available at that port.

image

### DIFF
--- a/bumps/webview/client/src/App.vue
+++ b/bumps/webview/client/src/App.vue
@@ -7,9 +7,10 @@ import {
   activePanel,
   addNotification,
   cancelNotification,
-  connected,
+  connecting,
   default_fitter,
   default_fitter_settings,
+  disconnected,
   file_menu_items,
   fileBrowser,
   FileBrowserSettings,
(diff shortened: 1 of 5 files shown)


pkienzle:
On browser update for a disconnected server tab in Chrome the `disconnected` box turns into `"Server Connecting..."` when the chrome update is complete. It reattaches when I restart the server (now conveniently on port 5148).

How about `Waiting for server...` without the quotes to cover both cases?

pkienzle:
I changed it to `Server connecting...` and `Server disconnected`.

pkienzle (approved):

PR #238, opened 2025-04-02
Ensure unique names for session history items

bmaranville:
This PR does 2 things 

1. Uses a dict for storing `HistoryItem` items in the session history, instead of a list
2. Uses a unique name derived from the timestamp instead of the raw timestamp as the key of that dict

This is to avoid a name conflict if more than one history item is appended within the resolution of the timestamp we are using (milliseconds)

### DIFF
--- a/bumps/webview/client/src/components/HistoryView.vue
+++ b/bumps/webview/client/src/components/HistoryView.vue
@@ -10,6 +10,7 @@ const props = defineProps<{
 }>();
 
 type HistoryItem = {
+  name: string;
   timestamp: string;
   label: string;
   chisq_str: string;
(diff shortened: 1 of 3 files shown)


bmaranville:
@pkienzle hopefully this will fix your problem where you ran into name collisions when writing session files during tests with simple models

bmaranville:
We were getting duplicate timestamps only on Windows platforms, and after looking into it the actual resolution of `datetime.datetime.now()` on Windows is more like 1.6 milliseconds even though it reports microseconds (while on linux/macos it really is microseconds)

So I think the approach in this PR (tagging new items with the same timestamp with a name e.g. <timestamp>-1, <timestamp>-2) should fix the problem, and we're only likely to see those duplicate timestamps on Windows.

PR #241, opened 2025-04-07
synchronize CLI fit options with the webview interface

bmaranville:
Populates the shared state in webview (`api.state.shared`) with all the default fit options, updated with any overrides specified on the command line e.g. when doing 
```sh
bumps model.py --fit=dream --steps=1234
```
... the number of steps shown in the fit options dialog in the webview gui will indeed be 1234.

### DIFF
--- a/bumps/webview/server/cli.py
+++ b/bumps/webview/server/cli.py
@@ -498,7 +498,8 @@ def interpret_fit_options(options: BumpsOptions):
     on_startup.append(lambda App: api.state.shared.set("selected_fitter", fitter_id))
     # TODO: send commandline options to the webview interface
     all_fit_options = api._get_fitter_defaults()
-    all_fit_options[fitter_id]["settings"].update(dict(fitopts))
+    if fitter_id in all_fit_options:
+        all_fit_options[fitter_id]["settings"].update(dict(fitopts))
     api.state.shared.fitter_settings = all_fit_options
 
     api.state.parallel = options.parallel
--- a/bumps/webview/server/cli.py
+++ b/bumps/webview/server/cli.py
@@ -486,20 +486,25 @@ def interpret_fit_options(options: BumpsOptions):
         TRACE_MEMORY = True
         api.TRACE_MEMORY = True
 
-    fitopts, errors = fit_options.update_options(options.fit_options)
+    fitopts, errors = fit_options.check_options(options.fit_options)
     if errors:
         warnings.warn("\n".join(errors))
     # TODO: leave fitter_id in fitopts
     # TODO: use dict rather than list of pairs for fitopts
     fitter_id = fitopts.pop("fit")
 
-    # on_startup.append(lambda App: publish('', 'local_file_path', Path().absolute().parts))
-    api.state.shared.selected_fitter = fitter_id
-    if isinstance(api.state.shared.fitter_settings, dict) and fitter_id in api.state.shared.fitter_settings:
-        api.state.shared.fitter_settings[fitter_id]["settings"].update(fitopts)
+    # Add secret fitter to the GUI (pt and scipy.leastsq as of this writing)
+    # TODO: FitOptions.vue needs to update fit_names on changeActiveFitter
+    if fitter_id not in api.state.shared.fitter_settings:
+        fitter = fit_options.lookup_fitter(fitter_id)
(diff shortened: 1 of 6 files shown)


bumps/webview/server/cli.py
```diff
@@ -497,6 +497,10 @@ def interpret_fit_options(options: BumpsOptions):
     # on_startup.append(lambda App: publish('', 'local_file_path', Path().absolute().parts))
     on_startup.append(lambda App: api.state.shared.set("selected_fitter", fitter_id))
     # TODO: send commandline options to the webview interface
+    all_fit_options = api._get_fitter_defaults()
+    if fitter_id in all_fit_options:
+        all_fit_options[fitter_id]["settings"].update(dict(fitopts))
+    api.state.shared.fitter_settings = all_fit_options
```
pkienzle, line 504: Both fitter_settings and fitter_id use shared state. It is odd that one is delayed until startup and the other set immediately.

Also, calling the "private" function `_get_fitter_defaults` from another module is discouraged. [Yes, I know that it is because `get_fitter_defaults` is an async function, so you just needed another name.]

Having the initialization of shared state sprinkled about will be problematic. For example, the jupyter startup does not go through the cli fit options processor, and so `fitter_settings` will not be initialized.

I suggest initializing state and shared state as much as possible when the state variable is created in api.py, and provide `api.set_fit_options(fitopts)` to update it with command line options. Since fitopts also contains `fit: fitter_id` this function can set shared.fitter_id as well.

The  `set_fit_options` function can call `fit_options.update_options` to verify the options. The `update_options` function should be renamed since it is not updating. I'm not sure what to do with the errors.

bmaranville: I was making a minimal change... in reality the delayed setting of "selected_fitter" could be done immediately as well, since all this is happening before the server is initialized (and there's no reason to wait until after the server is initialized). 

When the client connects it asks for all the values in `api.state.shared`

bmaranville: It seems awkward to set the fit options for e.g. "dream" by sending a dict with `{"fit": "dream", **actual_fitter_options}`.  Isn't it more straightforward to call a function like `set_fit_options(fitter_id: str, new_fit_options: dict)` instead of picking apart the payload dict after the fact?

bmaranville:
agreed that the Jupyter users will want to update the fit options as well, in a reasonable manner.

bumps/webview/server/cli.py
pkienzle: I believe it is slightly easier for the user to treat the fitter id as one of the fitting options when programming their fitting workflow in jupyter. For example:
```python
seq = [
    dict(fit='amoeba', radius=0.5),
    dict(fit='amoeba', radius=0.1),
    dict(fit='dream', init='eps', 'burn'=500, 'samples'=50000),
]
for opts in seq:
    await api.start_fit_thread(opts)
    await api.wait_for_fit_complete()
```

The current `start_fit_thread` interface would either require a sequence of tuples, or the awkward call:
```
await api.start_fit_thread(fitter_id=opts['fit'], options=opts.copy().pop('fit'))
```

I updated the `start_fit_thread` interface as part of the cli reconciliation PR but backed it out. Even though it is a simpler, it may not be enough to justify changing long-standing internal interfaces.

pkienzle:
> agreed that the Jupyter users will want to update the fit options as well, in a reasonable manner.

I think the jupyter code does call the command line startup code, but it feels like the wrong place to set the initial fit options shared state. I would still update the options there, but only for the fitter given on the command line.

bmaranville:
Now, `api.start_fit_thread(fitter_id: str, options: Optional[Dict[str, Any]])` uses the fitter settings stored in `api.state.shared.fitter_settings[fitter_id]` by default, falling back to the default fitter settings if the shared settings are not populated.

Also, `api.state.shared.fitter_settings` is populated with the fitter defaults as part of `webserver.setup_app`, before `cli.interpret_fit_options` is called, so that the shared settings can be updated with any fitopts from the command line.

When running in batch mode, the shared settings will not be populated, so the default settings will be used.

It's possible we could change it around so that `interpret_fit_options` returns fitopts and fitter_id in addition to startup and shutdown actions, so that `interpret_fit_options` would not be directly modifying the shared state as a side-effect.

pkienzle (approved):

PR #242, opened 2025-04-09
Add max time option to fitters

pkienzle:


bumps/webview/server/fit_options.py
```diff
@@ -199,36 +214,39 @@ def form_fit_options_associations():
...
         if key == "fit":
             # Already added.
             continue
-        if key not in defaults:
+        if key not in defaults and key != "time":
             # Skip unrecognized options
             unknown.append(f"{key}={value}")
             continue
-        stype = FIT_OPTIONS[key].stype
+        stype = float if key == "time" else FIT_OPTIONS[key].stype
```
bmaranville, line 249: If there is a need for special handling for the `time` kw that is injected into the fit options, would it make more sense not to inject it into the fit options for each fitter in the first place?  Especially since it's not being sent to the fitter classes?
Is it done this way so that the time option shows up in the fit options panel in the GUI?

PR #245, opened 2025-04-14
Show final step on console and rename near_best to jump

pkienzle:
Refactor the interaction between the fitters and controller.

MonitorRunner now includes info() for updates to the user during the fit and final() for updates at the end of the fit which are forwarded to each monitor.

MonitorRunner now includes stopping() so that the fitter can ask if the abort flag has been set. This replaces the abort_test parameter passed to each fitter. [It doesn't save much code, but it makes clear that MonitorRunner is handling bidirectional communication between the controller and the fitter.]

The fitters were all adapted to accept this interface. The internals of several fitters were already set up to look at the output of the monitor update to determine if the user requested stop. For these fitters, the redundant abort_test parameter was removed. [The abort_test interface is easier to understand; in retrospect I should have removed the test against monitor update return value.]

Adapting multistart fits to info() revealed some problems in the interface. In particular, that the user couldn't specify the jump distance. The `--near_best` option was renamed `--jump=radius`.

Testing the updates somehow showed that get01/put01 wasn't working for indefinite intervals. The fix is included in this patch. getfull/putfull was also tested and adjusted. There is still an unexpectedly high loss of precision for bounded intervals. I didn't fix it because we currently aren't using the getfull/putfull interfaces, and instead raise a "not implemented" error.

I updated the "random lines" and "particle swarm" fitters so they at least run, but they are too unreliable to include in the available fitter list.

Use the simplified fit interface in the doc build. fitters.fit() now accepts fit="fitter" as well as method="fitter" so that the dictionary produced by the argument parser can be used directly. [Note: refl1d/doc/sitedoc.py will require the same changes as in bumps]

### DIFF
--- a/bumps/bounds.py
+++ b/bumps/bounds.py
@@ -69,6 +69,7 @@ __all__ = [
     "SoftBounded",
 ]
 
+import sys
 from dataclasses import dataclass, field
 import math
 from math import log, log10, sqrt, pi, ceil, floor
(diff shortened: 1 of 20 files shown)


bumps/bounds.py
```diff
@@ -715,6 +728,7 @@ def __init__(self, mean: float = 0, std: float = 1, limits=(-inf, inf), hi="inf"
             # for backward compatibility:
             lo, hi = limits
         limits = (-inf if lo is None else float(lo), inf if hi is None else float(hi))
+        # TODO: Why not use dot notation? class Bounds doesn't define __setattr__
```
bmaranville, line 731: using `frozen=True` with dataclass sets trap on setattr, so you have to use `object.__setattr__` if you use a custom init with a frozen dataclass.

PR #249, opened 2025-04-24
Add store/export to simple fit interface

pkienzle:
Allows:
```python
result=fit(problem, name="basename", export="export/path", store="session.h5", resume=last_result)
```

Replaces #68 using the new internal interfaces.

### DIFF
--- a/bumps/fitters.py
+++ b/bumps/fitters.py
@@ -864,7 +864,7 @@ class DreamFit(FitBase):
 
         def update(state, pop, logp):
             # Get an early copy of the state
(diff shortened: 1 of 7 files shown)


bmaranville:
I think my only comment on this is that it seems pretty awkward to start a fit with the `MPMapper` enabled (have to initialize the mapper with the problem, and also pass the problem to the `fit()` call.)

I don't have a good sense of how hard it would be to use this function with `MPI` either.

bmaranville (approved):

PR #251, opened 2025-04-29
Client fit fields from server

bmaranville:
Currently the fitter fields (like steps, xtol, etc.) are defined on the server in `bumps.webview.server.fit_options`, and a matching definitions is maintained in TypeScript on the client in `bumps/webview/client/src/components/FitOptions.vue`

This PR

- adds a method that converts the FIT_OPTIONS from the server into a JSON-compatible form
- adds a method to the webview api for retrieving those fields
- refactors the FitOptions panel in the client to use these values
   - (including adding new tooltips to the labels, with the extended description!)
- removes the deprecate TS definition of FIT_FIELDS in the client code

### DIFF
--- a/bumps/webview/client/src/components/FitOptions.vue
+++ b/bumps/webview/client/src/components/FitOptions.vue
@@ -111,6 +111,13 @@ function validate(value: any, field_name: string) {
   }
   if (field_type === "integer" && parseInt(value, 10) != float_value) {
     return false;
+  } else if (typeof field_type === "object") {
+    if ("min" in field_type && float_value < field_type.min) {
+      return false;
+    }
+    if ("max" in field_type && float_value > field_type.max) {
+      return false;
+    }
   }
   return true;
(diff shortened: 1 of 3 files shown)


pkienzle (approved):

bmaranville:
It currently shows a red border with an exclamation mark if you put e.g. `jump radius = 5.0` in the fit options for `amoeba`... the only way to figure out what's wrong is to hover over the field and see that the description lists the range of [0,0.5]

I'm not sure what else to do with that... I could refactor it so that `validate()` returns a reason if it fails instead of true/false.

bmaranville:
Ok, I'm now calculating an error message in `validation_error(value...)` method instead of just returning true/false (it returns `null` if there is no validation error).

Hovering over the value when it is highlighted red and marked as "invalid" will show the reason it is invalid, e.g. `must be <= 0.5`

PR #253, opened 2025-04-30
Serialize curvefit with generic function pickling

pkienzle:


bmaranville:
I tried both methods of storing a serialized fitproblem (as bytes, and as string following the procedure in the PR), and the compression for some reason (for the test JSON data I used) didn't really work for the string data.  Maybe the default chunking is different for string dtypes or something, but I ended up with a 52 kB .h5 file for the string data, and a 15 kB .h5 file for the bytes data, both with compression=5

(The string being stored is 49 kB in size)

bmaranville:
I think compression can be enabled for string dtypes by also specifying the string length in the dtype (not for variable-length strings)

bumps/webview/server/state_hdf5_backed.py
```diff
@@ -47,7 +47,7 @@
 
 
 SESSION_FILE_NAME = "session.h5"
-ARRAY_COMPRESSION = COMPRESSION = 5
+ARRAY_COMPRESSION = COMPRESSION = 9
```
bmaranville, line 52: I don't know about increasing this... in some tests I've run the size with compression=9 is almost the same as with compression=5, but it takes significantly longer to write (sometimes many times longer)

Issue #260, opened 2025-05-07
DREAM fit cannot be resumed from GUI without reloading session


hoogerheide:
A DREAM fit was started from the web GUI with 10000 burn steps and 100 steps. After a full frame was finished (generation = 772 > 100 steps), the fit was stopped from the GUI and attempted to be resumed. The result was a failure of the assertion that the generation is equal to the number of steps.

The printed values above the traceback shows the various variables that are failing in the assertion.

![Image](https://github.com/user-attachments/assets/1e798c52-e43c-4b9c-ac32-486aad7c9313)

Note that if the following steps were taken, resuming works fine:
1. uncertainty state was saved to a session file
2. Refl1D was restarted
3. the session file was opened
4. the uncertainty state was reloaded

This seems to suggest that when the fit is complete, the internal values of `self.generation`, `self._update_count`, and `self._thin_count` are not the same as the values they assume after an uncertainty state is loaded from a session.

@bmaranville can comment on his ideas to simplify this by conditioning the uncertainty state after a fit is complete.

PR #266, opened 2025-05-15
improve dream resume support

pkienzle:
See #260

Fixes convergence test for --alpha in dream (the test was reversed).

Improves error reporting for webserver RPC calls.

Preserves total number of generations.

Note: DE reports wrong number of steps on resume, but that will be posted as a separate ticket.

### DIFF
--- a/bumps/dream/convergence.py
+++ b/bumps/dream/convergence.py
@@ -73,32 +73,32 @@ def ks_converged(state, trials=TRIALS, density=DENSITY, alpha=ALPHA, samples=SAM
     if alpha == 0.0:
         return False
 
-    # Make sure we have the desired number of draws
+    # Make sure we have a full buffer.
     if state.generation < state.Ngen:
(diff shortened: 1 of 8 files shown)


bumps/dream/convergence.py
```diff
@@ -73,32 +73,32 @@ def ks_converged(state, trials=TRIALS, density=DENSITY, alpha=ALPHA, samples=SAM
...
 
     n_draw = int(density * samples)
-    reject = _robust_ks_2samp(head, tail, n_draw, trials, alpha)
-    if reject:
-        return False
-
-    return True
+    p = _robust_ks_2samp(head, tail, n_draw, trials)
+    # print(f"convergence {state.generation}: pval={p} < {alpha}?")
+    return p < alpha
```
bmaranville, line 101: Has the user-facing definition of "Convergence (alpha)" changed because of this PR?

bmaranville: Also, since alpha is compared to a probability in [0,1], why do we limit it to [0,0.1]?

PR #269, opened 2025-05-20
Remove fixed and fittable attributes from ValueProtocol.

pkienzle:
slot.fixed and slot.fittable are never used. Remove them from ValueProtocol and its subclasses.

### DIFF
--- a/bumps/parameter.py
+++ b/bumps/parameter.py
@@ -518,7 +518,6 @@ class Parameter(ValueProtocol, SupportsPrior):
     def __repr__(self):
         return "Parameter(%s)" % self
 
-    # TODO: deprecate
     @classmethod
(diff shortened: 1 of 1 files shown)


bumps/parameter.py
```diff
@@ -330,7 +324,6 @@ class Parameter(ValueProtocol, SupportsPrior):
     # prior: Optional[BoundsType]
     id: str = field(metadata={"format": "uuid"})
     name: Optional[str] = field(default=None, init=False)
-    fixed: bool = True
```
bmaranville, line 333: we definitely want the the `fixed` attribute on `Parameter` - otherwise it will not be saved or restored.

pkienzle: oops

PR #271, opened 2025-05-21
Remove unused bumps.vfs module

bmaranville:
Removes a module with utilities for loading models from virtual filesystems, like from zip files.

If we keep the module, we should remove all the python 2.x compatibility shims and do other cleanup.


pkienzle (approved):

pkienzle:
Also needs to be removed from refl1d

bmaranville:
> Also needs to be removed from refl1d

I didn't think to check there!

PR #273, opened 2025-05-22
Improve parameter-linking copy functions

bmaranville:
avoid linking Calculation objects when making linked copies, add helper to make multiple copies

### DIFF
--- a/bumps/parameter.py
+++ b/bumps/parameter.py
@@ -23,7 +23,7 @@ import uuid
 from functools import wraps
 from enum import Enum
 
-from typing import Type, TypeVar, Optional, Any, Union, Dict, Callable, Tuple, List, Sequence
+from typing import runtime_checkable, Optional, Any, Union, Dict, Callable, Tuple, List, Sequence, Protocol
 from .util import Literal
 
 import numpy as np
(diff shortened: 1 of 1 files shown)


pkienzle:
Code is not used anywhere internally in bumps or refl1d. Does `copy_linked` appear in user models? Should we instead delete it?

hoogerheide:
Seems to me that this may be an important functionality for a model builder that generates multiple linked models. I would recommend keeping it for now.

pkienzle (approved):
I'm approving since the code works. However:

1. The unlinked parameters have the same name across models. This is confusing.

Perhaps add a suffix arg to the copy_linked function, and update cp.name with cp.name+suffix. The suffix should also be applied to the linked parameter so that it will have the unique name when it is later unlinked. Maybe want prefix and suffix, or just a template with a placeholder for the name.

2. Minor quibble about names:

* `has_parameter` sounds like a boolean; maybe `component` instead?
* `copy_linked` and `make_linked_copies` may be confusing.

3. `copy_linked` and `make_linked_copies` are not in bumps.names.

4. The contents of the slots are being deepcopy'd as well. This may lead to confusing semantics.

For example, if the original model has expressions involving external parameters it may be confusing that these parameters need to be listed in exclude.

Another example, I modified curvefit/curve.py example as follows:

```python
from bumps.parameter import copy_linked

M2 = copy_linked(M, exclude=[M.b])
M2.y += 1
M3 = copy_linked(M2, exclude=[M2.m])
M3.y += M3.x
```

Since M2.m is is linked to M.m, then there is not a separate slope parameter for the third model.

I can address the extra linkage by adding `M3.m.unlink()` statement and fitting works, though (4a) the new parameter is not visible to webview.

5. Should warn if a parameter listed in exclude is not in the model. This will help reduce user confusion in more complex cases.

PR #275, opened 2025-05-22
Use float32 for uncertainty state storage instead of float64

bmaranville:
This is a one-character PR, mostly here to invite discussion:

We can get much smaller stored versions of the dream state if we store them as float32 than float64, because in most cases for the models we're using we're not very sensitive to the extra precision, and in a random sampling scheme (like dream) the extra bytes are most likely uncompressible noise.

This translates into much better compression for float32. (worth collecting some numbers here)

### DIFF
--- a/bumps/dream/state.py
+++ b/bumps/dream/state.py
@@ -110,7 +110,7 @@ CREATE = gzip.open
 SelectionType = Optional[Dict[Union[int, Literal["logp"]], Tuple[float, float]]]
 
 
-UNCERTAINTY_DTYPE = "d"
+UNCERTAINTY_DTYPE = "f"
 MAX_LABEL_LENGTH = 1024
 LABEL_DTYPE = f"|S{MAX_LABEL_LENGTH}"
 H5_COMPRESSION = 5


pkienzle:
Given that this applies to the draw from the posterior, the only time it will be an issue is when 1-σ is smaller than about 0.001% of the parameter value. Then the uncertainty plots for the parameters will start to see strange spikes because the bin size will be finer than the least significant bit in the stored point values.

For consistency on resume it would be good if the current population were preserved separately in double precision even if the rest of the draw were chopped down to single precision. Similarly for best_x and best_logp so that χ² on the best doesn't jump on reload.

bumps/dream/state.py
```diff
@@ -128,6 +128,10 @@ def _h5_write_field(group: "Group", field: str, data: Union[NDArray, str]):
     elif isinstance(data, float):
         return group.create_dataset(field, data=data, dtype=np.double)
     else:
+        # Squash 64-bit arrays to 32-bit. This is lossy. The points in
+        # thin_draws may no longer evaluate to the values in thin_logp.
+        if data.dtype == np.float64:
+            data = np.asarray(data, dtype=UNCERTAINTY_DTYPE)
```
bmaranville, line 134: I think the hdf5 lib has very efficient casting routines so we don't need to do this.

pkienzle:
I revised it to use targeted 32-bit fields. Still haven't addressed the 64-bit current population problem.

pkienzle:
We may want to keep the state in 32-bit in memory as well, especially if we need to have all states loaded in memory at once during file rewrite. This would make behaviour completely consistent across reload. We can still preserve the double precision current population in the state so that the stepper preserves full precision.

pkienzle:
I've made new in-memory buffers for dream state 32-bit. On reload I preserve the type of the stored buffers. I've tested that I can start a new dream fit and resume it. I have not tested that I can resume from old hdf or mcmc files.

I don't think there is any need for storing samples with 64-bit since I'm adding 1e-6 relative noise at each step.

bumps/dream/state.py
```diff
@@ -204,24 +209,22 @@ class Fields: ...
     # print(f"loading {total_generations=} {Ngen=}")
     state._gen_offset = total_generations - Ngen
     state._gen_index = 0
-    state._gen_draws = Fields.gen_draws.astype(int)
```
bmaranville, line 198: these conversions were in here to support some older session files... in which the `_gen_draws`, `_thin_draws` and `update_draws` dtype was float

bmaranville: only 2 billion, since we're using a signed int ;)

PR #277, opened 2025-05-23
Set default alpha to zero for no convergence test

pkienzle:


bumps/webview/server/fit_options.py
```diff
@@ -92,7 +92,14 @@ def __repr__(self):
...
 Setting("xtol", "x tolerance", Range(0, 1), "Stop when population diameter < xtol relative to range.")
 Setting("ftol", "f(x) tolerance", float, "Stop when variation in log likelihood < ftol.")
-Setting("alpha", "Convergence", Range(0, 0.1), "Stop when probability that population is varying < alpha.")
+Setting(
+    "alpha",
+    "Convergence",
+    Range(0, 0.1),
+    """\
+    Stop when probability that population is varying < alpha or use default
+    zero for no convergence test.""",
```
bmaranville, line 101: Thanks for updating the help string!  I think this will really benefit the users.

PR #283, opened 2025-05-28
add chisq() method to FitProblem

### LINKED ISSUE #282: FitDriver.chisq() calls FitProblem.chisq(), which doesn't exist
GenX is using the `FitDriver.chisq` method to get the final value of chisq from a fit - but that method in turn calls `FitProblem.chisq`, which is not an implemented method.

See: https://github.com/bumps/bumps/blob/master/bumps/fitters.py#L1148

In the method `FitProblem.chisq_str`, we _do calculate chisq_, so should we just offer both interfaces in the API, and use `FitProblem.chisq()` in `FitProblem.chisq_str()`?

bmaranville:
fixes #282

### DIFF
--- a/bumps/bounds.py
+++ b/bumps/bounds.py
@@ -245,6 +245,10 @@ class Bounds:
     def limits(self):
         return (-inf, inf)
 
+    @property
+    def dof(self):
+        return 0
+
     def get01(self, x):
         """
         Convert value into [0,1] for optimizers which are bounds constrained.
@@ -607,10 +611,16 @@ class Distribution(Bounds):
     def __init__(self, dist):
         object.__setattr__(self, "dist", dist)
 
+    @property
+    def dof(self):
+        return 1
+
     def random(self, n=1, target=1.0):
         return self.dist.rvs(n)
 
     def nllf(self, value):
+        # TODO: This is inconsistent with normal below, which does not include normalizer
+        # What is the chisq/2 equivalent for an arbitrary distribution?
         return -log(self.dist.pdf(value))
 
     def residual(self, value):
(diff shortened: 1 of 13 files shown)


pkienzle (commented):
I dropped chisq from the interface because it is a confusing mess and people should not be tempted to use it.

Most existing models will do the right thing, however PoissonCurve uses Poisson statistics for the data, and the error-in-variables hack introduces a parameter-dependent rescaling of the error bars. Future modifications for nested sampling may want nllf to include the constant scale factor, Σ log(2πσ²)/2.

We should be reporting "corrected" parameter uncertainties, scaled by √(χ^2/dof). An obvious way to do this would be to call FitProblem.chisq(norm=True). In that case, chisq should include the prior probability, pparameter, in the result (For gaussian priors, each prior is equivalent to an additional data point and an additional fit parameter, so no change in dof.)

Some of the nllf methods in bounds.py include the scale factor. These should not be included when computing the "corrected" uncertainties.

The pconstraint factor is supposed to be a bowl guiding the fit to the feasible region. If it is working correctly the final result should not contribute to the nllf, so we can ignore it in the discussion of chisq. [Since it is unused in the final result, using nllf/dof for parameter uncertainty correction is reasonable, with the same caveat about scale factors in the priors.]

The chisq_str method lumps pparameter with pconstraint, only using pmodel for the displayed chisq value. This is equivalent to the chisq method in the PR, but it'll be inconsistent if the chisq method were to include parameter priors.

The server api get_chisq function returns 2*nllf/dof, where nllf is pparameter + pmodel + pconstraint. This may or may not be consistent, depending on how we implement FitProblem.chisq.

The chisq function for the individual fitness will have a different value for dof, so yet another way for inconsistencies to appear.

Make sure that genx is maximizing nllf, not chisq. Otherwise it won't respect constraints and priors. This may be a non-issue if genx doesn't use constraints and priors in its models.

Regarding approval I'm ambivalent. The PR makes things marginally worse by including yet another code path to maintain, and it may be misleading to the reader.

Note: copy these comments into the code base for future maintainers to read.

PR #288, opened 2025-06-06
Fix logic for automatic burn point determination

bmaranville:
When `--trim` is enabled as a fit option, DreamFit will try to determine an automatic "burn point" with a sliding-window test similar to the alpha convergence test.

The logic for this test was broken in some recent refactoring, this PR fixes the logic.

At the same time, the `--trim` option is removed from the interface right now because although this PR fixes the logic for finding an automatic burn point, that point is not stored in the fit_state and so is not available at the time plots are generated, so enabling `trim` has no effect on the export output or any of the webview plots.

### DIFF
--- a/bumps/dream/convergence.py
+++ b/bumps/dream/convergence.py
@@ -166,13 +166,14 @@ def burn_point(state, method="window", trials=TRIALS, **kwargs):
     return index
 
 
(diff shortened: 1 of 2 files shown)


bmaranville:
Closing in favor of #289

PR #289, opened 2025-06-09
Store user-controlled 'portion' setting in Dream state

bmaranville:
Instead of generating an ephemeral private `._trimmed` attribute on the fitter class when the `trim` option is enabled, store the calculated portion in a new attribute `portion` on the Dream state.

Then this `portion` is used by default in the various dream plots and other outputs.

It is a public attribute of the Dream state because the user may wish to set it manually as well, e.g. in a GUI. 

The stored portion is also used by default in any `state.draw()` unless a float (non-None) `portion` kwarg is supplied e.g. `state.draw(portion=0.5...)`


pkienzle (approved):

pkienzle:
Fixes #287

PR #290, opened 2025-06-09
Convergence show burn and control cutoff with slider

bmaranville:
This PR introduces enhancements to the convergence plot in the webview

- Adds a slider to control cutoff fraction (previously fixed at 0.25)
- Shows burned portion of samples in a different color, when the fitter supports this (dream)

image


bmaranville:
Refactored to show a line instead of trying to shade the 20% to 80% band of the burned region (much easier to implement and maintain)

image

bmaranville:
... now with labels
image

pkienzle (approved):

bmaranville:
Also added a control for the trim_portion, that is only shown at the end of a fit where the `portion` is defined (after a dream fit).

Allows manual adjustment of the dream `fit_state.portion`, which persists in saved sessions, and is used when calculating the uncertainty plots.

image

acaruana2009 (approved):

acaruana2009:
Happy with this overall. I do think we are going to need to do some explaining of when to use these features though - as I am getting a little confused to as to when you would use trim over just moving the the cut off for the burn in.

I am assuming it is working as: you set your desired samples at the start of the fit, with your burn in. After the fit the burn in line is just shown at these values specified. If you think the fit has not quite converged, you can sacrifice a few samples by either: adjusting the cut off or adjusting the trim portion? I feel like I am missing something here in its operation?

bmaranville:
The cutoff only affects the display of the convergence plot... It is set to 0.25 by default so that you don't get swamped on the y-scale by the earlier chisq values

acaruana2009:
To be clear, I think it is a useful control, however it is implemented - I am often trying to look at the convergence that it is at now rather than as you say the very large chisq at the start.

bmaranville:
Making the trim portion line update as you drag is certainly possible, though quite a bit of work to do it client-side since we're generating the plots on the server. The current behavior is just to send a new value for portion to the server when you stop dragging and then a new plot is requested from the server, which is quite robust.  We could add a second boolean argument to the server API function for setting portion so that it doesn't trigger an uncertainty plot update until you stop dragging, and only updates the convergence plot (a little janky, but similar to how the "cutoff" slider currently works)

I would like to keep the "burn" line on the graph though I'm happy to change the name.  I would also really like to name the sliders and graph lines such that they provide the most direct feedback to the user as to what they are looking at and what the controls do.  Since they are directly controlling the "portion" with the slider, I think we need to indicate that.  Having the "full portion" or "buffer" or "burn" line helps them quickly reason about what the slider actually does, and what the limit of 1.0 means.  I agree that using "standard" terms to describe the control elements would be ideal, but in this case they're not directly controlling the "burn-in", only indirectly.  It would be useful to add documentation and maybe tooltips that indicate how (portion) combined with (burn setting from fit options) combined with (total steps currently taken) contribute to the final effective "burn-in".

For the slider currently named "cutoff" - I don't understand how we can call that "steps" - since it is a number from 0 to 1 and doesn't control the number of steps shown even indirectly (number of steps changes with each update).  It seems like we could come up with some algorithm for automatically adjusting the cutoff based on the y-values on the plot to show progress, rather than just cutting off a fraction of the steps on the x-axis.  Maybe we should name the slider "steps cutoff fraction" for now until we figure out something better?

Agreed that a button to manually trigger an auto-trim would be a useful feature.  Also a thinning control of some sort.  Maybe we can make a new PR for those?

pkienzle:
Works for mac. I fixed the correlations/uncertainty plot freezes when the trim is set to 1.0 by keeping the last two generations.

pkienzle:
`stored_portion` error appears to be resolved.

The webview client test is failing during CI:
```
Run bun run test:types
  bun run test:types
  shell: /usr/bin/bash -e {0}
$ vue-tsc --noEmit --skipLibCheck -p tsconfig.json --composite false
error TS2688: Cannot find type definition file for 'node'.
  The file is in the program because:
    Entry point of type library 'node' specified in compilerOptions
```

This appears to be bit rot. It worked fine locally until I removed package-lock.json and rebuilt the client. Unfortunately I did not keep the old package-lock.json, so I'm not sure what changed.

Here's what I should have done:
```
cp bumps/webview/client/package-lock.json saved-package-lock.json
rm -r bumps/webview/client/{package-lock.json,dist,node}
python -m bumps.webview.build_client
cd bumps/webview/client
node_modules/.bin/vue-tsc --noEmit --skipLibCheck -p tsconfig.json --composite false
```
giving the error:
```
error TS2688: Cannot find type definition file for 'node'.
  The file is in the program because:
    Entry point of type library 'node' specified in compilerOptions

  tsconfig.json:17:15
    17     "types": ["node"],
                     ~~~~~~
    File is entry point of type library specified here.


Found 1 error.
```

bmaranville:
The profile plot redrawing bug was a separate issue - that panel was not responding to `updated_uncertainty` signals correctly.  It will be fixed in https://github.com/reflectometry/refl1d/pull/320 - there is no change needed to this PR.

PR #292, opened 2025-06-16
Fix method for starting webview server in jupyter

bmaranville:
This PR adds a method with the same name as the new one provided in refl1d PR 299 for starting a webview server within a jupyter notebook: `start_jupyter_server`

This method automatically enables the webview link display on server start (the option `jupyter_link=True` is passed to `start_app`), and it returns an asyncio task that should be awaited in the notebook, in order to capture any exceptions that are raised during the server start, e.g.
```python
from bumps.webview.server.webserver import start_jupyter_server

await start_jupyter_server()
```

### DIFF
--- a/bumps/webview/server/webserver.py
+++ b/bumps/webview/server/webserver.py
@@ -259,11 +259,12 @@ def start_from_cli(options: BumpsOptions):
 server_task = None
 
 
-def bumps_server():
+def start_jupyter_server():
     global server_task
 
     # Start the server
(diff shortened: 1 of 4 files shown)


bumps/webview/server/webserver.py
```diff
@@ -268,14 +268,16 @@ def bumps_server():
 
 
 async def start_app(
-    options: BumpsOptions,
```
bmaranville, line 271: The existing API before the CLI merge allowed calling start_app without any arguments.  This restores that functionality for existing notebooks.

PR #294, opened 2025-06-23
Fix parsing of model serialization in client

bmaranville:
Model serialization is returned as string now - update client to parse either string or bytes.

Fixes error shown in client anytime user selects model tab.


pkienzle (approved):
Code looks fine, and it fixes the model rendering bug.

With the client so tightly coupled to the server does it need to handle both string and bytes payloads? If we change back to a bytes representation can we do the transformation to string on the server and keep the client simple?

I don't understand why the linter is now giving errors. Maybe there was a recent change in the linting package?

bmaranville:
Fair 

> Code looks fine, and it fixes the model rendering bug.
> 
> With the client so tightly coupled to the server does it need to handle both string and bytes payloads? If we change back to a bytes representation can we do the transformation to string on the server and keep the client simple?
> 
> I don't understand why the linter is now giving errors. Maybe there was a recent change in the linting package?

Fair enough... simplifying.

PR #302, opened 2025-06-28
Fix script args for curvefit/poisson.py and test_functions/model.py

pkienzle:
Script args now working again. Dill was getting confused when trying to deserialize, forcing a reload of the original script.

I can't explain why it is working again. It is still be pickling a function defined in the script file, and so it should need more information to deserialize. Maybe it is because I converted PoissonCurve and PDF to dataclasses.

I refactored the code for relative imports from the script file. It now uses a temporary import hook to intercept imports from dependent modules. I removed the old relative_import support since it wasn't working.

The new code works for a single fit session but not for resume from history so for now I've disabled it. I have code to capture the context, but no way to get it into the ProblemState object.

Tested using:
```sh
$ rm /tmp/ex.h5; bumps --fit=dream --store=/tmp/ex.h5 doc/examples/curvefit/poisson.py --args poisson -b; bumps --store=/tmp/ex.h5 --fit=dream --resume -s
$ rm /tmp/ex.h5; (cd doc/examples/test_functions && PYTHONPATH=. bumps --fit=dream --store=/tmp/ex.h5 model.py --args gauss 3 -b; PYTHONPATH=. bumps --fit=dream --store=/tmp/ex.h5 --resume)
```

### DIFF
--- a/bumps/curve.py
+++ b/bumps/curve.py
@@ -521,6 +521,7 @@ def logfactorial(n):
     return result
 
 
+@dataclass(init=False, eq=False)
 class PoissonCurve(Curve):
     r"""
     Model a measurement with Poisson uncertainty.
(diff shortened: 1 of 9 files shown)


pkienzle:
With cloudpickle we can register imports as pickle by value rather than pickle by reference. We may be able to support relative imports from scripts again if we register them all when they are loaded.

https://github.com/cloudpipe/cloudpickle?tab=readme-ov-file#overriding-pickles-serialization-mechanism-for-importable-constructs

PR #318, opened 2025-08-15
Run the model uncertainty plugin during export

pkienzle:
In the 0.x version of bumps the model uncertainty plugin was run after a dream fit was completed, which created the profile uncertainty plots in the store directory.

This PR enables this behaviour during export from webview and from the command line.


bmaranville:
Superseded by #317 which rebased onto this PR and pulled in all its changes.

PR #320, opened 2025-08-18
Fix shutdown dialog

bmaranville:
pass disconnected state to Shutdown dialog as prop instead of trying to react to ref (which wasn't working)


pkienzle (approved):
Both the old and the new version work on mac safari

Issue #312, opened 2025-08-11
Need export button for custom plots


hoogerheide:
A user has requested a "data export" feature for custom plots.

This should be easy to do: update `CustomWebviewPlot` with an `exportdata` field containing CSV data. If this field is not empty, the custom plot panel will show an "Export" button that will download the CSV data.

Advantages:
- Simplicity -- responsibility for generating export data is with the custom plot function that creates the `CustomWebviewPlot` object
- Keeps CSV generation on the server side

Disadvantages:
- The same data are being sent twice, once in plot form and once in CSV form
- This is particularly egregious for "table" type custom plots where the plotdata is already in CSV form. Perhaps some special logic could be implemented here. On the other hand, so far the tables haven't been that big.

pkienzle:
Custom view data should be exported along with all the other data.

We should also have the option of storing it in the session file in case it is expensive to recalculate.

The profile uncertainty plot should be able to be rendered as a custom plot (#314).

hoogerheide:
> The profile uncertainty plot should be able to be rendered as a custom plot

We don't currently support non-generic controls in the custom plots. The profile uncertainty plot is a bit different because it has a selection box for multiple profiles. This will have to be addressed if we want to move it into the custom plot universe.

PR #323, opened 2025-08-21
Export data from custom plots

hoogerheide:
Adds an export data button to the custom plot views, if export data is provided.

Addresses https://github.com/bumps/bumps/issues/312

Two notes:
- to test using `molgroups` one will need to install the appropriate `molgroups` branch: https://github.com/reflectometry/molgroups/tree/export_custom_data
- The export data button doesn't always appear in the right place in the Custom Uncertainty tab. Perhaps someone with more bootstrap experience could try to amend this.

### DIFF
--- a/bumps/webview/client/src/components/CustomPlot.vue
+++ b/bumps/webview/client/src/components/CustomPlot.vue
@@ -9,6 +9,8 @@ import type { AsyncSocket } from "../asyncSocket";
 import { configWithSVGDownloadButton } from "../plotly_extras";
 import { setupDrawLoop } from "../setupDrawLoop";
 
+const hidden_download = ref<HTMLAnchorElement>();
+
 type PlotInfo = { title: string; change_with: string; model_index: number };
 // const title = "Custom";
 const figtype = ref<string>("");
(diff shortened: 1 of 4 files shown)


hoogerheide:
> Some questions:
> 
> (1) Should every view have an export button? (2) Should the export file menu trigger export for every view? (3) Should custom views be able to include data to the session file?

(1) It should only be visible if export data is actually available, as in this PR; but it's an intriguing idea. Many of the tabs would benefit from this (for example, exporting the summary tab would be a nice way of getting a table of fit parameters and ranges for a publication).

Re (2) and (3): Both would work best if each view had only one thing in it. The custom plots can have any number of "sub-views"; only those that are explicitly selected by a user are calculated and are cached somewhere. So any other export trigger, or something that writes to the session file, would either have to calculate these (potentially time-consuming and expensive) or only include those that are already cached. Conversely, reloading a state from the session file would have to reconstruct the cache. I don't have a strong opinion about this -- my applications using molgroups calculate quickly enough (even on a laptop) to generally not be too painful to recalculate.

hoogerheide:
> * The export data button doesn't always appear in the right place in the Custom Uncertainty tab. Perhaps someone with more bootstrap experience could try to amend this.

This is fixed by https://github.com/bumps/bumps/pull/323/commits/12452197425950a8b720d931b10cdb9825506bdf

pkienzle (approved):
Code looks fine and works as advertised. I merged in master so that the tests will pass. No conflicts to resolve.

pkienzle:
Apparently still some lint fixes to be done...

PR #324, opened 2025-08-23
Create freevars with independent parameters

pkienzle:
The copy semantics for parameters gave each parameter its own ID, but left the slots the same.

This had the effect of tying the values of the parameters together, though not in the same way as `p1.equals(p2)`, so bumps was unable to identify that the parameters were tied.

This PR also updates `unlink` so that neither Calculation nor Constant set by the model can be overridden.

Tested manually under normal usage conditions with `sasmodels/examples/simul_fit.py`. A minimal set of automated tests would be nice.

Refactored handing of model_update() so that models will clear caches when cycling through the free parameters.

### DIFF
--- a/bumps/webview/server/api.py
+++ b/bumps/webview/server/api.py
@@ -1408,12 +1408,12 @@ def params_to_list(params, lookup=None, path="", freevars=None) -> List[ParamInf
         else:
             value_str = VALUE_FORMAT.format(nice(params.value))
             has_prior = getattr(params, "prior", None) is not None
-
-            writable = has_slot and isinstance(params.slot, (Variable, Parameter))
+            writable = has_slot and isinstance(params.slot, (float, Variable, Parameter))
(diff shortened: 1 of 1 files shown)


bumps/parameter.py
```diff
@@ -706,10 +706,15 @@ def remove_tag(self, tag: Optional[str] = None):
...
-        """copy will only be called when a new instance is desired, with a different id"""
+        """
+        Copy will only be called when a new instance is desired, with a different id.
+        Make sure the slot is not shared with the original, otherwise the parameters
+        will be tied together after the copy.
+        """
         obj = type(self).__new__(self.__class__)
         obj.__dict__.update(self.__dict__)
         obj.id = str(uuid.uuid4())
+        obj.slot = copy(self.slot)
```
bmaranville, line 717: So for a `Variable` instance in the slot, this will result in an unlinked copy (because the value attribute is a plain float), but for a `Parameter` in the slot this will trigger a recursive call to this same `__copy__` method on that instance - is that the desired behavior?

bmaranville (approved):

PR #326, opened 2025-08-27
Fix check for bounds in add_prior with normal distribution

pkienzle:
Trivial fix to bug discovered by inspection. Did not test.

### DIFF
--- a/bumps/parameter.py
+++ b/bumps/parameter.py
@@ -177,6 +177,12 @@ class Calculation(ValueProtocol):  # the name Function is taken (though deprecat
 
 
 class SupportsPrior:
+    """
+    Parameter mixin allowing the value to be restricted to a prior probability.
+
+    Every prior has hard limits and bounds.
+    """
+
     prior: Optional[BoundsType]
     limits: Tuple[float, float]
     distribution: DistributionType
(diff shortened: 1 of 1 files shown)


bumps/parameter.py
```diff
@@ -198,6 +204,10 @@ def add_prior(
         bounds: Optional[BoundsType] = None,
         limits: Optional[Tuple[float, float]] = None,
     ):
+        # TODO: why are bounds and limits not being updated in the underlying object?
+        # Note: limits should come from the model (i.e., the parameter object)
+        # and bounds should come from the user (i.e., the slot value)
```
bmaranville, line 209: I don't understand this note - the bounds and limits are both attached to the parameter object, and applied to whatever's in the slot that provides a value.

pkienzle: The system is confusing. We have both bounds and limits and they can exist in both the parameter itself and in the parameter that fills the slot.

What we want are hard constraints (e.g., length >= 0) imposed by the model on any value that occupies that position in the model, as well as user defined priors on that particular value (e.g., length ~ N(μ=100, σ²=10)) which comes from information about the system from other sources.

So indeed my comment is wrong: both (model) limits and (user) bounds belong to the model "slot" and apply to any value/variable/expression/calculation/constant that happens to occupy it.

The "TODO" refers to the usage of add_prior. The only place it is called with bounds and limits is in model_reset, and there it is just passing in the bounds and limits that are already present, so no need to update. The other place it is called is in the api after an update to the parameter fit range. The short term fix is to remove bounds and limits from the add_prior function.

The same parameter can have multiple constraints from the various roles it occupies in the models being fit. We need to verify that all of these are satisfied when doing a simultaneous fit. Similarly, if an expression occupies the model slot we need to verify that the constraint is satisfied for the expression value.

PR #327, opened 2025-08-29
Tests get python versions from pyproject

bmaranville:
in this PR, the test CI for github is updated to use a script that reads `pyproject.toml` and parses the `requires-python` key to get a minimum and maximum allowed python version.

Tests will fail unless both a minimum and maximum version are defined here with `>=min_version,<=max_version` (order can be reversed, but can't use `>` or `<` or `==`)
Every minor version of python from min to max (inclusive of endpoints) will be tested in `test-publish.yml`.

**NOTE** the minimum version of python is used for testing with the `macos-latest` and `windows-latest` os runners, and is also used for testing the docs build on `ubuntu-latest`

An alternative test matrix, that would be easier to run, would be to test just the minimum and maximum supported python versions, and maybe to test against the maximum version for the other two os runners?


pkienzle (approved):

glass-ships (approved):
Note that 3.9 reaches end of life next month: https://devguide.python.org/versions 
so we may want to consider pushing min up to 3.10. 
Otherwise looks good! i've never been motivated enough to try writing a custom github action in JS so it's interesting to see (none of the functions are called, only exported - does it just run the default export when it gets used via `needs`? or

bmaranville:
The only thing that gives me pause about merging this is that it sets an upper limit on the python version, which seems artificial.  On the one hand, it means you can only install it for python versions that have passed the tests (good!) but then it's very likely that python will change in backward-compatible ways, so that it would keep working with newer versions.

This will be fine, as long as we keep releasing new versions of this package that support the newer versions of python!

bmaranville:
Drop this in favor of #431

PR #330, opened 2025-08-29
Modify dream interfaces to make plotting easier from scripts.

pkienzle:


bumps/dream/state.py
```diff
@@ -190,7 +250,8 @@ class Fields: ...
     portion = getattr(Fields, "portion", 1.0)
 
     # Create empty draw and fill it with loaded data
-    state = MCMCDraw(0, 0, 0, 0, 0, 0, thinning, portion=portion)
+    state = MCMCDraw(0, 0, 0, 0, 0, 0, thinning)
+    state.portion = portion
```
bmaranville: This is just from curiosity... what is the difference in effect from calling the constructor `MCMCDraw(..., portion=portion)` vs. setting the portion attribute after initializing the instance?

PR #333, opened 2025-09-02
make legend draggable on convergence plot

bmaranville:
@acaruana2009 noticed that the burn line(s) on the convergence plot sometimes overlap the legend now.

This PR makes the legend draggable through plotly configuration, so the user can at least manually move things into a non-overlapping state.

It's possible we should also move the labels on the burn lines to be closer to the midpoint, so that this overlap is less likely overall?


bmaranville:
This is a minimal, manually tested change with no impact outside the convergence panel in the webview GUI.  Merging without review.

PR #342, opened 2025-09-16
Improve plotting support for models in webview

pkienzle:
Improve interactions with problems, results and sessions in jupyter notebooks.

Allow plotly plots in fitness classes.

Added support for `python -m bumps.nested model.py` to run nested sampling via ultranest. This doesn't belong with the rest of the changes, but it is too difficult for end users to pull together multiple PRs into their python environment, so I'm bundling it all together.

Details:

* matplotlib style sheet for readable labels in webview
* add zoom and mouse position plugins for interactive plotting
* figure size chosen to have reasonable graph margins, though perhaps only for my monitor 8-(
* push_model() context for plotting a single model
* move χ² to the top of the plot (mpl, plotly, no plotter, exported plots)
* show χ² for each model as well as overall χ²
* use plotly if model defines a plotly() method (takes precedence over matplotlib)
* added convergence to the simple fitter results, and added plot_convergence(results) to plot it (mpl)
* added reload_state_from_export("model.par") and reload_fit_from_session("session.h5")
* rearranged code so that std/cov are attached to the problem rather than the fit driver (which doesn't exist on reload)

Define an interface for reading/writing session/export/server from a jupyter notebook

Bumps functions:
```python
import bumps.names as bp
bp.help()                                            # display this help
problem = bp.load_problem(path, args=[arg1, ...])    # load model from script (.py) or export (.json)
options = dict(fit=dream, burn=100)                  # fit options (see bumps -h for list)
!bumps --help                                        # show available options
fitresult = bp.fit(problem, **options)               # synchronous fit interface
fitresult = bp.fit(problem, resume=fitresult, **options) # resume a fit from fitresult
bp.plot_convergence(fitresult)                       # show convergence plot
bp.show_results(problem, fitresult)                  # summarize fit results
problem.plot()                                       # show model plots
fitresult.state.show()                               # show dream plots
problem, fitresult = bp.load_fit_from_session(path)  # load bumps MCMC from session file
problem, fitresult = bp.load_fit_from_export(path)   # load bumps MCMC files that were exported
```

Webview functions:
```python
await bp.start_bumps()                               # start the webview server
bp.display_bumps(height=600)                         # diplay webview in jupyter
await bp.set_problem(problem)                        # send problem to webview
await bp.start_fit_thread(options=options)           # start the webview fit thread
problem, fitresult = await bp.get_fit_from_webview(wait=True) # get fit results from webview
await bp.set_problem(problem, fit=fitresult)         # send problem and results to webview
```

MCMC statistics: [TODO]
```python
# parameter summary table
# range restriction and brushing
# calling individual plots
# derived and hidden parameters
# entropy calculation
```

### DIFF
--- a/.github/workflows/test-publish.yml
+++ b/.github/workflows/test-publish.yml
@@ -64,7 +64,7 @@ jobs:
       matrix:
         cfg:
           #- { os: ubuntu-latest, py: 3.8 }
-          - { os: ubuntu-latest, py: "3.9" }
+          #- { os: ubuntu-latest, py: "3.9" }
           - { os: ubuntu-latest, py: "3.10" }
           - { os: ubuntu-latest, py: "3.11", doc: 1 }
           - { os: ubuntu-latest, py: "3.12" }


bmaranville:
I would like to have something more abstracted for plotting... matplotlib is hard-coded into many corners of the code and it would be nice to ask for a plot and specify a backend instead of doing a search through class methods.

bumps/webview/server/webserver.py
```diff
@@ -303,13 +319,18 @@ async def start_app(
         enable_convergence_kernel_heartbeat()
 
     if jupyter_link:
-        return open_tab_link()
+        open_tab_link()
```
bmaranville: Does this work in jupyter notebooks?  I think the return statement was here to make sure it displays the link as the output of the cell.

bumps/webview/server/api.py
```diff
@@ -71,9 +99,9 @@ async def to_thread(func, *args, **kwargs):
 MODEL_EXT = ".json"
 TRACE_MEMORY = False
 
-
+# TODO: reloading the module wipes out state
```
bmaranville: What is the use case for reloading the module?  Don't you have to try hard to do that?

bumps/webview/server/api.py
```diff
@@ -127,84 +155,139 @@ async def emit(
...
+    if not filename and problem_path:
+        filename = Path(problem_path).name
+    if not path and problem_path:
+        path = Path(problem_path).parent
+    name = name or problem_name or filename
     state.shared.updated_model = now_string()
     state.shared.updated_parameters = now_string()
     state.shared.custom_plots_available = get_custom_plots_available(problem)
-    # invalidate the uncertainty state:
-    state.reset_fitstate()
```
bmaranville: Doesn't this show the old fit result while a new fit is running?  Isn't that confusing?

bumps/fitters.py
```diff
@@ -1390,59 +1356,311 @@ def register(fitter, active=True):
...
+    problem.plot(fignum=2)
+    if results.state:
+        results.state.show()
+    else:
+        problem.show_err(results.x, results.dx)
+
+
+# Note: get fit is currently computing the covariance matrix each time the fit result is
+# retrieved. This will go away if it is saved along with the webview fit results.
+async def get_fit_from_webview(wait: bool = False):
```
bmaranville: Why are these webview functions defined here in fitters.py?

pkienzle:
> Do the separate PRs still exist in a usable form, as they were merged to make this big one? It may be easier for an end user to try out the result, but it is much harder for a reviewer to figure out which parts might break core functionality and which parts are safe to apply immediately.

The original PRs still exist. My plan was to review the PRs individually and address any issues in the merged PR.

bumps/webview/server/webserver.py
bmaranville: I don't know that there are any existing notebooks to protect.

PR #354, opened 2025-10-11
Threadpool mapper (for free-threaded python)

bmaranville:
This PR adds a new mapper based on `ThreadPoolExecutor`, for fitters that use a mapper (like `dream`)

Each thread gets a copy of the fitproblem, which is updated on `start_mapper` (when a new fit is started)

**NOTE**
This will only be really useful with free-threaded python interpreters (e.g. python 3.14), and should have a lot less memory overhead than `multiprocessing`

Quick tests indicate that it is 5% to 10% faster than `MPMapper` under the same conditions, too (using `c_ext` backend for refl1d, since `numba` is not available yet for free-threaded builds)

Issues in `numba` repo suggest that they will have support for free-threaded python soon (no show-stoppers, code is mostly thread-safe already)

Tracking dependencies needed for this to work:
* h5py/h5py/issues/2475
* numba/llvmlite/issues/1338
* numba/numba/issues/10303


bumps/webview/server/cli.py
```diff
@@ -386,6 +386,11 @@ def get_commandline_options(arg_defaults: Optional[Dict] = None):
         type=int,
         help="run fit using multiprocessing for parallelism; use --parallel=0 for all cpus",
     )
+    misc.add_argument(
+        "--threads",
+        action="store_true",
+        help="run fit using multithreading for parallelism",
+    )
```
pkienzle: Given that this is replacing MPMapper I would prefer that this happen implicitly if the environment supports it rather than having to explain to users when they should or shouldn't use it.

pkienzle (approved):
Untested but code looks good.

Note that `multiprocessing.cpu_count()` returns the number of processors on the node, not the number of cpus available to the task. On linux we should use `len(os.sched_getaffinity(0))`. This may be different if you are running on slurm with `--cpus-per-task` less than the number of cpus on the node. If we do this then we don't have to remember to use `--parallel=$SLURM_CPUS_PER_TASK` on the bumps command line in our batch script. The same applies for MPMapper.

MacOS and Windows do not support `os.sched_getaffinity`, so fall back to `multiprocessing.cpu_count`.

Note also that we can't mix MPI and multiprocessing in slurm. The task affinity for the individual mpi tasks is set to a single core (one of pair of virtual cpus on intel), so multiprocessing dumps the entire pool onto that core. Not that this matters for bumps since we can only use MPIMapper, MPMapper or ThreadMapper.

PR #356, opened 2025-10-30
[draft] Switch to biome for frontend linting/formatting

glass-ships:
[BiomeJS](biomejs.dev/) is a linter and formatter written in rust that provides:

- Significantly faster linting (upwards of 10x, annecdotally up to ~50x faster than eslint)
- 97% compatibility score with prettier
- Simpler configuration, fewer files (one json file, schema well documented)
- Has IDE support for VSCode, Zed, and other editors
- Provides [git-hooks](https://biomejs.dev/recipes/git-hooks/), if we so desire

Will follow up with a secondary PR to address some lingering linting errors.

Decided it was probably better to start with bumps rather than refl1d (see https://github.com/reflectometry/refl1d/pull/342).


bmaranville:
The tool is still not ready for linting, I think... I tried it after merging master back in, and it can't see that functions defined in <script setup> in a Vue component are used in the template (marks them as unused functions, suggests removing them or prepending with underscore).

Maybe we just use biome to replace prettier?

bmaranville:
It does seem like they're working on it though... https://github.com/biomejs/biome/pull/9283

bmaranville:
Closing in favor of #434 - will revisit linting later

PR #361, opened 2025-11-05
Add button to toggle light/dark theme

glass-ships:
The only real changes here are `components/ThemeToggle.vue` and using it in `App.vue`.

The rest are linting/formatting changes from pre-commit, which seems like it probably hasn't run in a while. 

As for the failing frontend types test, I'm not immediately sure why it's failing in CI but passing locally and on other branches like #360 but I'll look into it


bmaranville (commented):
@pkienzle rejected an earlier button for toggling dark mode, in favor of a solution that inherits the mode from the OS or browser, though this new button is different in the sense that it overrides the inherited value instead.

The original counter-argument was that it cluttered the interface, and people who care about dark mode have already enabled it on their system.

What was the motivation for this new button?  A user request?

PR #364, opened 2025-11-18
Move freevars functionality to subclass

bmaranville:
This PR moves all freevars functionality to a subclass.

This achieves:

- 99% use case (no freevars) now has simple list of models instead of generator

**NOT TESTED**
(it passes existing pytest tests, but likely will fail in real use without some more work)

**TODO**:
if this seems like a reasonable thing to do, we can add a migration so that existing serialized models with non-trivial freevars will be converted to the new class on load.

### DIFF
--- a/bumps/fitproblem.py
+++ b/bumps/fitproblem.py
@@ -51,6 +51,7 @@ import logging
 import os
 import sys
 import traceback
+from types import NoneType
 from typing import Generic, TypeVar, Union, Optional
 import warnings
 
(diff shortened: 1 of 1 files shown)


pkienzle:
> In almost (all?) uses of `FitProblem.models` internally, it is just treated as an iterator (`for model in problem.models...`), so those uses could just be updated to accept types `FitProblem | FreeVarsFitProblem`. The major exception seems to be in the presentation of parameters, which currently checks the "freevars" attribute - this should be refactored to be handled internally in the `FreeVarsFitProblem.parameters()` function

The issue is not with the typing. The issue is with the semantic interpretation of problem.models[k], which has to set the parameters and clear the fitness caches when freevars is active. If we set problem.models to a special list-like container which does this that may be good enough. Code which uses it still needs to be careful, though. It is no problem if we say `Mk = model[k]; Mk.method1(); Mk.method2();` but horribly inefficient if we say `model[k].method1(); model[k].method2();`.

So that gets back to the use case that is asking for this feature. Why does it want to access a single model as a list instead of running through all models via a generator? For those cases, can't it just use model = problem.set_active_model(k)? Yes, that is a funky name, but the underly operation is funky so it is a good warning to read the docs.

Or perhaps better:
```python
with problem.push_model(k) as model:
    ...
```

The point is, until we eliminate freevars anyone who is playing with the models in the problem will need to be careful.

PR #369, opened 2025-11-21
Add max_steps calculator to all fitter classes

bmaranville:
For some fitters, calculating the maximum number of fitting steps from the input options is non-trivial (dream)

This PR adds a `classmethod` to all fit classes for calculating max_steps, with this signature (problem arg is required in `DreamFit.max_steps`, to get number of fitted parameters)
```python
@classmethod
def max_steps(cls, problem: Optional["bumps.fitproblem.FitProblem"] = None, options: Optional[Dict[str, Any]] = None) -> int:
```

It also then uses this method in the webview api when starting a fit, to populate the number of expected steps in the current fit.

### DIFF
--- a/.github/workflows/test-publish.yml
+++ b/.github/workflows/test-publish.yml
@@ -29,7 +29,7 @@ jobs:
           path: bumps/dream/random123
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v7
 
       - name: Set up Python
         uses: actions/setup-python@v6
@@ -77,7 +77,7 @@ jobs:
       - uses: actions/checkout@v6
 
       - name: Install uv
-        uses: astral-sh/setup-uv@v6
+        uses: astral-sh/setup-uv@v7
 
       - name: Set up Python ${{ matrix.cfg.py }}
         uses: actions/setup-python@v6


bumps/fitters.py
```diff
@@ -1004,6 +1031,20 @@ def error_plot(self, figfile):
...
+        options = _fill_defaults(options, cls.settings)
+        num_fitparams = len(problem.getp())
+        steps = options["steps"]
+        if steps == 0:  # Steps not specified; using samples instead
+            pop, draws = options["pop"], options["samples"]
+            pop_size = int(np.ceil(pop * num_fitparams)) if pop > 0 else int(-pop)
+            steps = (draws + pop_size - 1) // pop_size
+        steps += options["burn"]  # add burn-in steps
+        DE_steps = Dream.DE_steps  # DreamFit does not override DE_steps, uses default
+        steps = int(np.ceil((steps - 1) / DE_steps)) * DE_steps + 1  # round up to multiple of DE_steps
```
bmaranville: I don't understand this line.  Specifically, why the value of `(steps -1)` should be rounded to the nearest multiple of `DE_steps` and then have 1 added back to it.  I determined this algorithm empirically by trying combinations of `burn` and `steps` inputs with a `dream` fit.

Any ideas, @pkienzle ?

pkienzle (approved):
My inclination is to merge this PR, and save the refactoring of state for a later PR.

PR #371, opened 2025-11-24
Improve interactive plotting of Curve class in webview (plotly)

bmaranville:
Adds a plotly() method to `bumps.curve.Curve` class, providing better interactive plotting in webview interface.


bmaranville:
@pkienzle - why was this closed without merging?

PR #372, opened 2025-11-24
Fix showing multiple data plots in webview

bmaranville:
Currently the Data panel in webview has a toggle for showing multiple plots simultaneously, but then no way to actually show all those plots.

In this PR: 

- the api method on the server is modified so that it returns a list of figures, one per model_index in the `model_indices: List[int]` argument
- the client makes a grid of plots, one per selected model index.


bmaranville:
image

PR #380, opened 2025-12-04
Remove constraints=function from FitProblem interface.

pkienzle:
The constraints function interface to FitProblem grabbed the parameter objects from the global context to check if the values were within the feasible region. The function was then serialized and sent to the fit worker processes for evaluation. On deserialization the parameters referenced in the function no longer corresponded to the parameters in the model, and so the constraints were never applied for the point being evaluated.

This PR removes the constraints function from the interface, and changes the example to use an inequality constraint instead. The example now runs.

In order for the fit to guide a parameter into the feasible region, the constraint penalty must be a finite value plus the squared distance to the boundary. The penalty value must be large enough that points outside the boundary have a higher nllf than points inside the boundary, but small enough that penalty plus distance is different from penalty for small distances.

Test using bumps/doc/examples/constraints/inequality.py, which requires b > m but starts with b < m:
```sh
$ bumps inequality.py -s --fit=amoeba -b --session=/tmp/T1.h5
$ bumps inequality.py -s --fit=de -b --session=/tmp/T1.h5
$ bumps inequality.py -s --fit=dream -b --session=/tmp/T1.h5
$ bumps inequality.py -s --fit=lm -b --session=/tmp/T1.h5
```

Note: lm is failing, giving a fitted point outside the feasible region. This is true even when it is started with b > m. Fixing this is outside the scope of this PR.

### DIFF
--- a/bumps/fitproblem.py
+++ b/bumps/fitproblem.py
@@ -588,7 +588,14 @@ class FitProblem(Generic[FitnessType], CovarianceMixin):
         # Note: pmodel is zero if any constraints are failing. In that case the cost
         # will be the squared distance from the boundary of the feasible region for
         # the breaking parameters so that gradient descent can guide us back to the
-        # feasible region. The penalty is a large constant offset so that
+        # feasible region. The penalty would ideally be greater than any value of
+        # pmodel inside the boundary (otherwise the fitter may prefer the point in the
+        # infeasible region) but it has to be small enough that adding a small distance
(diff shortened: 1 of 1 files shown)


bmaranville (approved):
Fix incomplete documentation... otherwise good to go!

pkienzle:
I updated the comment to complete the thought. I agree that constraint handling is broken.

Issue #363, opened 2025-11-17
Move jupyter and scripting support code into its own module


pkienzle:
The fitters.py module includes jupyter and scripting support code that does not belong there. Move it to its own module.

Remove any circularity amongst `{fitters,fitproblem}.py` and `webview/server/{api,cli,fit_options,fit_thread,webserver}.py`.

Issue #22, opened 2019-04-17
consistent handling of cov/stderr/chisq


pkienzle:
`bumps.fitters.FitDriver` and `bumps.fitproblem.FitProblem` implement cov, stderr and chisq in a variety of ways.   Make sure they all go through the same code so that the results are consistent.

pkienzle:
`FitProblem.chisq` and `FitProblem.chisq_str` use different methods.

bmaranville:
Remember we (mostly you) addressed some of these issues in the major refactor in the dataclass_overlay branch (a good portion of the commits in that branch are unrelated to the serializable structure and are just refactoring/bugfixes)

pkienzle:
For independent Gaussian measurements nllf = Σ (y-f)²/ 2σ² + Σ ln (2πσ²)/2 = χ²/2 + C. The complete nllf also includes prior parameter probability and constraints, but these do not contribute to chisq.

* `fitproblem.FitProblem.chisq` uses `Fitness.chisq` if it exists otherwise it returns Σ R² / DOF where R = `Fitness.residuals()` if that exists, otherwise it raises an error. This function is mostly not used, and it fails for MultiFitProblem.
* `fitproblem.FitProblem.chisq_str` uses nllf/DOF where nllf comes from the fitness function. nllf may or may not include the normalizing constant  Σ ln (2πσ²) / 2. Curve.nllf does not.
* `pymcfit.PyMCProblem.chisq` returns nllf = - Σ ln P`. No normalization by DOF. This includes the normalization constant so it will be Σ ln (2πσ²)/2 more than χ²/2 unnormalized.
* `pymcfit.PyMCProblem.chisq_str` uses `%g` to format the return value from `PyMCProblem.chisq`. No uncertainty.
* `pdfwrapper.PDF.chisq` returns `nllf / DOF`. If nllf is computed correctly then this will be off by a scaling constant Σ ln (2πσ²) / DOF.
* `pdfwrapper.PDF.chisq_str` uses `%g` to format `chisq`. No uncertainty.
* `fitters.FitDriver.chisq` returns `problem.chisq`. It needs to be the normalized χ² / DOF.  It is used rescale the parameter uncertainties when measurement uncertainty is not provided. It is not defined for MultiFitProblem.
[...]

PR #385, opened 2025-12-09
Add --cov --err and --entropy cli so examples/entropy will run

pkienzle:
Add --cov --err and --entropy command line parameters similar to those available in bumps pre-1.0. This is needed for doc/examples/entropy.py.

This code will need to be cleaned up as part of the #363 refactoring.

Pre-1.0 bumps used FitDriver to compute the covariance matrix because it was sometimes provided by the underlying optimizer. For example, L-M calculates the Jacobian at every step, and BFGS keeps a running estimate of the Hessian.

In webview the FitDriver object is only available in the fit thread, so we don't have access to it at the end of the fit command. Instead we compute it from the numerical derivative (all but dream), or generate the equivalent statistics from the posterior sample (dream). The simple fit interface (fitters.fit) still uses the old FitDriver interface, but on problem reload from session or export the FitDriver is again unavailable and it reverts to the numerical derivative. This may give slightly different results to the uncertainties depending on how we got there.

The current PR prints the summary objects at the end of the fit using _show_results() in webview/cli.py, but uses fitters.show_table() to print the results at the end of fitters.fit(). To use show_table() in cli we would need to create an OptimizeResult object. We would also need show_cov() and show_entropy(), or extend show_table to output cov and entropy. We should harmonize the OptimizeResult object returned from fitters.fit() with the fit state object held in the webview state.fit_state object so we have consistent handling of stderr and cov (#22).

### DIFF
--- a/bumps/webview/server/cli.py
+++ b/bumps/webview/server/cli.py
@@ -74,7 +74,9 @@ class BumpsOptions:
     # Fitter controls.
     fit_options: Dict[str, Any] = field(default_factory=dict)
     resume: bool = False
-    # fit_outputs: Dict[str, Any] = field(default_factory=dict)
+    show_cov: bool = False
+    show_err: bool = False
+    show_entropy: Optional[str] = None
 
     # Session file controls.
     session: Optional[str] = None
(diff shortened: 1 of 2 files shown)


bmaranville:
Impressive, linking to what is now the oldest open issue in bumps...

PR #387, opened 2025-12-10
Minor fixes to type hints in bumps.dream.convergence

bmaranville:
Convert docstring type hints to new annotation style, and fill in missing type hints for arguments to functions.

### DIFF
--- a/bumps/dream/convergence.py
+++ b/bumps/dream/convergence.py
@@ -100,6 +100,8 @@ def ks_converged(
     n_draw = int(density * samples)
     p = _robust_ks_2samp(head, tail, n_draw, trials)
     # print(f"convergence {state.generation}: pval={p} < {alpha}?")
+
+    # comparison returns np.bool type which is not exactly the same as bool
     return bool(p < alpha)
 
 


bumps/dream/convergence.py
```diff
@@ -99,10 +100,10 @@ def ks_converged(state, trials=TRIALS, density=DENSITY, alpha=ALPHA, samples=SAM
     n_draw = int(density * samples)
     p = _robust_ks_2samp(head, tail, n_draw, trials)
     # print(f"convergence {state.generation}: pval={p} < {alpha}?")
-    return p < alpha
+    return bool(p < alpha)
```
pkienzle: Does the `bool()` do anything here?

If so, it is probably hiding a bug somewhere.

bumps/dream/convergence.py
```diff
@@ -137,9 +138,7 @@ def _check_nllf_distribution(data, df, n_draw, trials, alpha):
     return alpha > np.mean(p_vals)
 
 
-def burn_point(state, method="window", trials=TRIALS, **kwargs):
-    # type: ("MCMCDraw", str, int, **dict) -> int
-
+def burn_point(state: "MCMCDraw", method="window", trials: int = TRIALS, **kwargs) -> int:
```
pkienzle: `method: str = "window"`

pkienzle: It looks like we put it there in beginning in anticipation of more methods to come:
    https://github.com/bumps/bumps/blame/58ff3f4f40870ddd7471fee0351bed7b63018636/bumps/dream/convergence.py

I don't mind if it goes away.

PR #397, opened 2026-01-13
Fix initpop cov

### LINKED ISSUE #396: Covariance initializer for dream fails in batch file
On updating to 1.0.3 from 0.9.3 I am no longer able to use the covariance initializer described [here](https://bumps.readthedocs.io/en/stable/guide/optimizer.html#fit-dream).

The following are the options I have used:
`python -m bumps.cli TestLine.py --fit=dream --trim=true --burn=350 --samples=1.5e5 --outliers=iqr --checkpoint=%checkpoint% --batch --parallel=0 --init=cov --store="%startLocation%\%saveFolder%\%saveFile%"`

Where TestLine.py is a model I use for debugging and is fitting y=m*x+b to 5 randomly placed points. Before upgrading I had no issue using the options listed. The error message is this:

```
Traceback (most recent call last):
   File "<frozen runpy>", line 198, in _run_module_as_main
   File "<frozen runpy>", line 88, in _run_code
   File "C:\Users\vboxuser\miniconda3\Lib\site-packages\bumps\cli.py", line 731, in <module>
      main()
   File "C:\Users\vboxuser\miniconda3\Lib\site-packages\bumps\cli.py", line 703, in main
      best, fbest = fitdriver.fit(resume=resume_path)
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "C:\Users\vboxuser\miniconda3\Lib\site-packages\bumps\fitters.py", line 1119, in fit
      x, fx = fitter.solve(
              ^^^^^^^^^^^^^
   File "C:\Users\vboxuser\miniconda3\Lib\site-packages\bumps\fitters.py", line 888, in solve
      population = initpop.generate(self.problem, **options)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   File "C:\Users\vboxuser\miniconda3\Lib\site-packages\bumps\initpop.py", line 56, in generate
      cov = problem.cov()
            ^^^^^^^^^^^
AttributeError: 'FitProblem' object has no attribute 'cov'
```

Running with the default initializer does not have any issue.

bmaranville:
User was not able to run cli with initpop=cov like before.  

This PR passes the needed parameter vector `x` to `FitProblem.cov(x)` in `bumps.initpop.generate`

Also handles `FitProblem` objects without a defined `name` attribute in cli fits.

fixes #396

### DIFF
--- a/bumps/cli.py
+++ b/bumps/cli.py
@@ -215,7 +215,7 @@ def make_store(problem, opts, exists_handler):
     if getattr(problem, "store", None) is None:
         raise RuntimeError("Need to specify '--store=path' on command line or problem.store='path' in definition file.")
     stem = (
-        problem.path.stem
+        Path(problem.path).stem
         if hasattr(problem, "path")
         else sanitize_filename(problem.name)
         if problem.name


bumps/cli.py
```diff
@@ -210,7 +210,8 @@ def make_store(problem, opts, exists_handler):
         problem.store = opts.store
     if getattr(problem, "store", None) is None:
         raise RuntimeError("Need to specify '--store=path' on command line or problem.store='path' in definition file.")
-    problem.output_path = os.path.join(problem.store, problem.name)
+    problem_name = problem.name if problem.name else "problem"
```
pkienzle: Problem name used to default to script name, but this condition is no longer being triggered because problem.name is initialized to None if there is no problem name:
https://github.com/bumps/bumps/blob/e7ce10d3e6526d20d19a445f3c37268463c0a023/bumps/fitproblem.py#L1031-L1032
This should change to:
```python
if not getattr(problem, "name", None):
    problem.name = path.stem
```

We still need to fix the code here in case the problem wasn't loaded from a script.

Changing the semantics a bit, we can use the script name, but default to the problem name or to "problem" if there is no script. Sanitize the filename to protect against non-filename characters in problem.name, such as `name = "Ni/Fe multilayer"`.
```python
import re
def sanitize_filename(name: str) -> str:
    return re.sub(r'[^A-Za-z0-9._-]', '_', name)
...
stem = problem.path.stem if hasattr(problem, "path") else sanitize_filename(problem.name) if problem.name else "problem"
problem.output_path = os.path.join(problem.store, stem)
```

bumps/initpop.py
```diff
@@ -53,7 +56,7 @@ def generate(problem, init="eps", pop=10, use_point=True, **options):
     if init == "random":
         population = random_init(pop_size, initial, bounds, use_point=use_point, problem=problem)
     elif init == "cov":
-        cov = problem.cov()
+        cov = problem.cov(initial)
```
pkienzle: The reported issue #396 says that problem is missing the cov attribute. This change does not address the issue.

Note that `problem.cov()` is equivalent to `problem.cov(problem.getp())`. Because initial is set to `problem.getp()` this change has almost no effect (initial has non-finite values replaced by 1.0, so there is an effect on poorly defined problems).

PR #412, opened 2026-02-17
Fix MPI session start errors (remove autosave from load_problem)

bmaranville:
remove `state.autosave()` from `set_problem`, add to `load_problem_file` and `set_serialized_problem`

Should address the same issues as #411 but without adding complexity for tracking MPI rank in the state.

### DIFF
--- a/bumps/webview/server/cli.py
+++ b/bumps/webview/server/cli.py
@@ -537,6 +537,7 @@ def interpret_fit_options(options: BumpsOptions):
     write_session = options.write_session if options.write_session is not None else options.session
 
     # The logic for setting the current fit problem is complex.
+    #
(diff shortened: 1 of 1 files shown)


bumps/webview/server/api.py
```diff
@@ -308,7 +310,6 @@ async def set_problem(
         if state.shared.autosave_history and state.problem is not None and state.problem.fitProblem is not None:
             await save_to_history(f"Loaded model: {name}", keep=True)
         await add_notification(content=path_string, title="Model loaded", timeout=2000)
-    state.autosave()
```
bmaranville: was this one causing problems? It seemed like a natural place to save, at the time (on fit completion?) Is there another trigger that overlaps?

pkienzle: This is the tail end of `set_problem()`. The fit complete handler (end of fit) and the uncertainty update monitor (checkpoint interval) still call autosave.

PR #422, opened 2026-03-11
include examples in bumps install

### LINKED ISSUE #383: Link in Poisson docs requires auth
The [documentation](https://bumps.readthedocs.io/en/stable/tutorial/curvefit/poisson.html) for "Fitting Poisson data" contains a [link](https://www-cdf.fnal.gov/physics/statistics/notes/pois_eb.txt) to "the notes from the CDF Statistics Committee". However this link requires authentification with Fermilab. 

Is this intended? Or am I missing something? 

Image

pkienzle:
For convenience when teaching bumps or testing an installation it is useful to have the tutorial examples available in the installed package.

This PR puts the examples into the package so that they get installed.

It adds bumps.examples() to return a path to the examples.

Leaving as draft until we solve the following:

* how run an example from the command line
* how to load an example in webview
* how to load an example into jupyter
* how to handle examples when bumps is part of a bundled application
* how to load into webview running through pyodide

Fixes #383


bumps/__init__.py
```diff
@@ -20,6 +20,17 @@
 __schema_version__ = "1"
 
 
+def examples():
```
bmaranville: Seems like this will be more intuitive if named e.g. `get_examples_path`, so you can do 
```python
from bumps import get_examples_path
from bumps.fitproblem import load_problem

examples_path = get_examples_path()
curve_example = load_problem(examples_path / "curvefit" / "curve.py")
```

bmaranville (approved):
This seems like a straightforward change - any reason not to include it in the next release?

PR #423, opened 2026-03-13
reenable soft_range support

pkienzle:


bmaranville (commented):
Can we add any tests that will check that this is functioning as intended?

PR #426, opened 2026-04-15
Centralize bumps api

bmaranville:
Currently, the active cli module for bumps (the one called if you type "bumps" or "python -m bumps" at the command line) is located at `bumps.webview.server.cli`, and it is importing the state and api etc. from that parent package (`bumps.webview.server`), which is awkward.  Also, modules like `bumps.fitters` were importing from `bumps.webview.server.state_hdf5_backed` etc.

This PR moves all the business logic for a stateful interaction with the bumps library to the main `bumps` package level.  These modules are moved:

- `bumps.webview.server.api` -> `bumps.api`
- `bumps.webview.server.fit_thread` -> `bumps.fit_thread`
- `bumps.webview.server.logger` - > `bumps.logger`
- `bumps.webview.server.state_hdf5_backed.py` -> `bumps.state`
- `bumps.cli` -> `bumps.cli_old` (retained during transition)
- `bumps.webview.server.cli` -> `bumps.cli`

In addition, a new `bumps.plots` package was created, and these plotting modules were moved:

- `bumps.webview.server.corrplot` -> `bumps.plots.corrplot`
- `bumps.webview.server.varplot` -> `bumps.plots.varplot`
- `bumps.webview.server.convergence_plot` -> `bumps.plots.convergence_plot`
- `bumps.webview.server.errplot` -> `bumps.plots.errplot` (this one is likely to be deprecated soon)

As the old `bumps.cli` was renamed and fully deprecated, `bumps.cli.load_best()` function was moved so it is now a method of `bumps.fitproblem.FitProblem` classes.  All uses of the previously deprecated `bumps.cli.load_model` were updated to use `bumps.fitproblem.load_problem` directly

Lastly the entry point was updated to use `bumps.cli:main` directly.


bumps/fitproblem.py
```diff
@@ -962,18 +963,74 @@ def plot(self, p=None, fignum=1, figfile=None, view=None, model_indices=None):
...
+            values.append(default_value if is_empty else remaining.pop(0))
+            undefined.append(is_empty)
+        self.setp(np.asarray(values))
+        if any(undefined):
+            self.undefined = np.asarray(undefined)
+
+        # Note: restore default behaviour of getstate/setstate rather than
+        # inheriting from BaseFitProblem
+        def __getstate__(self):
+            return self.__dict__
```
pkienzle, line 1023: Too much indent

bumps/cli.py
```diff
@@ -1,714 +1,1150 @@
+# Note: the following text appears at the end of the bumps -h command line help.
```
pkienzle, line 1: Using `git mv` might have preserved history  better, and made it easier to see what has changed from webview/server/cli.py + cli.py to cli.py + cli_old.py.

Maybe it would have been better to leave the old cli.py in place and call the new one something different.

Probably too late now.

bumps/plots/corrplot.py
pkienzle, line 1: Some of these plots are specific to dream.

Maybe move dream/plotly/corrplot.py, with dream/corrplot.py moving to dream/mpl/corrplot.py ?

Maybe extract the Corr2D data structure to dream/stats.py and hand it to the corrplot function?

These steps could be done as a separate plotting cleanup PR.

bumps/webview/server/webserver.py
pkienzle, line 1: Move to bumps/webview/server.py and remove the server subdirectory (which now contains only __init__,py and webserver.py)

bumps/api.py
```diff
@@ -64,8 +64,8 @@
 import bumps.errplot
 from bumps.util import push_mpl_backend
 
-from . import fit_options
-from .state_hdf5_backed import (
+from bumps import fit_options
+from bumps.state import (
```
pkienzle, line 68: Can now use relative imports from bumps

bumps/fit_options.py
pkienzle, line 1: confusion between options and fit_options. Maybe move options.py to gui/options.py and old_cli.py to gui/cli.py. Not sure what other cruft we can abandon.

bumps/fitters.py
```diff
@@ -1773,8 +1773,8 @@ def fit(
     from pathlib import Path
     from scipy.optimize import OptimizeResult
     from .mapper import MPMapper, SerialMapper
-    from .webview.server.fit_thread import ConvergenceMonitor
-    from .webview.server.state_hdf5_backed import now_string
+    from .fit_thread import ConvergenceMonitor
```
pkienzle, line 1776: ConvergenceMonitor needs to move into fitters.py. Or even better, be implicit in FitDriver since it is now expected for every fit.

PR #434, opened 2026-05-01
Biomejs formatter

bmaranville:
Use biomejs for formatting instead of prettier.  Biome is more robust and requires very little configuration.

Replaces #356 for now (maybe can use biome for linting too, in the future)


bmaranville:
@glass-ships I think this is a good first step... using biome just for formatting, and eslint just for linting. 

prettier is giving me all sorts of problems and it's complicated to troubleshoot - I like the simplicity of biome and it seems to work great as a formatter.

(It looks like they're working hard on the linting too, so soon we could probably switch to just biome for linting + formatting)

glass-ships:
i think i agree with that yeah - it's tough to find a better linter that knows how to handle Vue <script> sections and stuff, but at least for now it's a good formatter and we can look at switching things up in the future as it develops

PR #435, opened 2026-05-02
move active code out of bumps.cli so it can be deprecated

pkienzle:
Redoing #426 in stages.

First move all of the active functions out of bumps/cli.py so that it can be moved or removed.

Update all references to the new locations.

Import the functions back into cli.py so that third party packages can be updated asynchronously.

Also includes step 2: restructuring of option handling so that bumps/options.py can be moved or removed.

### DIFF
--- a/bumps/fitproblem.py
+++ b/bumps/fitproblem.py
@@ -31,10 +31,10 @@ Summary of problem attributes::
     plot(figfile: str, view: str) -> None
 
     # Set/used by bumps.cli
-    model_reset() -> None # called by load_model
-    path: str  # set by load_model
-    name: str # set by load_model
-    title: str = filename # set by load_moel
+    model_reset() -> None # called by load_problem
+    path: str  # set by load_problem
+    name: str # set by load_problem
+    title: str = filename # set by load_problem
     options: List[str]  # from sys.argv[1:]
     undefined:List[int]  # when loading a save .par file, these parameters weren't defined
     store: str # set by make_store
@@ -49,7 +49,7 @@ __all__ = ["Fitness", "FitProblem", "CovarianceMixin", "load_problem"]
 from contextlib import contextmanager
 from dataclasses import dataclass
 import logging
-import os
+import re
 import sys
 import traceback
 from typing import Generic, TypeVar, Union, Optional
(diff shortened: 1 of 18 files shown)


bumps/fitproblem.py
```diff
@@ -971,6 +971,72 @@ def __setstate__(self, state):
...
+
+def dump_pars(problem, path: Path, filename: str):
+    # TODO: Include non-fitted parameters in .par file?
+    # TODO: replaces api._write_pars
+    pardata = "".join(f"{name} {value:.15g}\n" for name, value in zip(problem.labels(), problem.getp()))
+    with open(path / filename, "wt") as fd:
+        fd.write(pardata)
+
+
+def load_pars(problem, path):
```
bmaranville, line 985: This seems like a pretty clear candidate to be a method on FitProblem... when is the first argument ever not a FItProblem instance?

bumps/webview/server/api.py
```diff
@@ -1537,7 +1534,14 @@ async def get_dirlisting(pathlist: Optional[List[str]] = None):
 
 @register
 async def get_fitter_defaults():
-    return fit_options.FITTER_DEFAULTS
+    # TODO: Need a better way to include experimental fitters in the web ui.
```
bmaranville, line 1537: This refactor of the fitter defaults and how they interact with the webview display seems unrelated to the bumps.cli refactor - can it be done separately?  What is the goal of these changes?

bmaranville: I think I was wrong - it looks like there is a tie-in where fit_options is being refactored slightly too, but the goal is still unclear to me.

bmaranville (approved):

bumps/webview/server/api.py
pkienzle: I was preparing to move options.py along with cli.py when I noticed that SasView depends on it.

After refactoring a bunch in preparation for upgrading SasView, I realized that I can continue to use the existing options handler by pulling in a couple of old symbols. We still want to upgrade SasView to use the new options handler but now we can do it after the bumps release.
 
Putting it in the same PR rather than stacking made reviewing the parts more complicated. Sorry!

PR #436, opened 2026-05-06
Centralize bumps api reprise

pkienzle:
Reprise of #426, done in stages.

Step 3: move cli+options to the gui subpackage and raise cli+options from webview.

Step 4: move support modules out of webview (merged from #437)

This is a breaking change since bumps.cli.main() is replaced by the webview equivalent. The old main() is moved to bumps.gui.old_cli.main(). The corresponding PR will be posted for refl1d.

Tested configurations:

    {bumps, refl1d} X  {cli, old cli, webview, wx}
    sasview
    ipython tutorial notebook (not yet published)

Other symbols are loaded from old_cli to bumps.cli and from old_options to bumps.options so that SasView can delay the transition to the new fit options representation.


pkienzle:
That's annoying... the github diff for this PR doesn't recognize that cli,options were moved to gui and replaced by cli,fit_options from webview/server. Because of #435 these files should only have changes to the import statements, but it would be nice of github would show this.

bmaranville (approved):

PR #441, opened 2026-06-03
Fix fitter export

bmaranville:
@mdoucet reported that export from `bumps.fitter.fit` was not working for `dream` fits.

Minimal reproduction:
```python
from bumps.fitters import fit
from bumps.fitproblem import load_problem

p = load_problem("doc/examples/curvefit/curve.py")
result = fit(p, method="dream", burn=100, steps=1000, export="try-export")
```

This PR addresses two separate issues:

1. Previously the `fit` function called `export_fit` with `result.state` in the 'fit` argument.  `export_fit` looks for an attribute `fit.state`, so we need to pass `result` directly
2. The converter from numpy to json-compatible types was only handling `ndarray`, but scalars and booleans were being sent to the function as well during export.

Should be tested that it doesn't mess up sasview (@krzywon)

### DIFF
--- a/bumps/dream/stats.py
+++ b/bumps/dream/stats.py
@@ -223,10 +223,16 @@ def numpy_json(o):
     To automatically convert numpy data to lists when writing a datastream
     use json.dumps(object, default=numpy_json).
     """
-    try:
+    if isinstance(o, np.integer):
+        return int(o)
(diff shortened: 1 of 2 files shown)


pkienzle (approved):
Code looks good. Example runs.

krzywon (approved):

PR #443, opened 2026-06-06
feat(webview): improve file browser drive and parent navigation

pquarterman17:
Navigation improvements for the webview file browser, developed and daily-driven on the refl1d side (as an overridden component) and now contributed upstream where they belong.

### 1. Drive list works on Python 3.10/3.11 (backend)

`get_dirlisting` populates `drives` via `os.listdrives()`, which only exists on **Python 3.12+**. Bumps supports **Python ≥3.10**, so Windows users on 3.10/3.11 silently get `drives = []` and the drives bar never appears. Added a `ctypes` `GetLogicalDrives()` fallback (same output, all supported versions). Non-Windows platforms still return `[]`, so the drives bar remains a Windows-only element.

### 2. Parent-directory navigation (frontend)

Currently the only way to go *up* is clicking an earlier segment of the breadcrumb. Added the two conventions users expect from file managers:

- an **⬆ button** at the front of the breadcrumb, and
- a **`..` entry** pinned at the top of the Subdirectories list.

Both go up one level and are hidden at a drive root.

### 3. Drives bar visibility + labeling (frontend)

The drives row was only shown for `drives.length > 1`; now shown for ≥1 (it doubles as a "jump to drive root" shortcut, useful even with a single drive). Added a bold **"Drives:"** label so the row of bare `C:\ D:\` links is self-explanatory, and fixed its `aria-label` — it was tagged `"breadcrumb"`, colliding with the actual breadcrumb nav below it.

### 4. Leftover debug line removed (frontend)

```html
<div>show files: {{ settings?.show_files }}</div>
```

rendered literally ("show files: true") in the dialog for every user.

## Notes

- **Stacked on #442** (`fix/network-drive-stall`) — the backend change lands inside the `_get_dirlisting_sync` helper introduced there, so this PR includes that commit. Happy to rebase onto master if #442 is declined or needs changes.
- Any of the UX pieces (2/3) can be dropped or tweaked if you have different preferences — they're independent.

## Testing

- `py_compile` clean; `_get_drives()` smoke-tested on Windows: `os.listdrives()` branch and the `ctypes` bitmask branch return the identical drive list.
- Client `biome format`, `eslint`, and `vue-tsc` checks pass on the modified component.
- The equivalent changes have been in daily use in a refl1d fork (Windows, multi-drive + network shares).

### DIFF
--- a/bumps/webview/server/api.py
+++ b/bumps/webview/server/api.py
@@ -1498,30 +1498,32 @@ async def get_topic_messages(topic: Optional[TopicNameType] = None, max_num=None
         return list(itertools.islice(q, start, q_length))
 
 
-@register
-async def get_dirlisting(pathlist: Optional[List[str]] = None):
-    # GET request
-    # TODO: use psutil to get disk listing as well?
+DIRLISTING_TIMEOUT = 10.0  # seconds before an unreachable path returns an error instead of hanging
+
+
+def _get_dirlisting_sync(pathlist: Optional[List[str]] = None):
+    """Walk a directory synchronously; run via asyncio.to_thread (filesystem I/O may be slow on network drives)."""
     subfolders = []
     files = []
     path = Path(state.base_path) if (pathlist is None or len(pathlist) == 0) else Path(*pathlist)
+    missing = None
     if not path.exists():
-        await add_notification(
-            f"Path does not exist: {path}, falling back to current working directory",
-            title="Error",
-            timeout=2000,
-        )
+        missing = path
         path = Path.cwd()
 
     abs_path = path.absolute()
     for p in abs_path.iterdir():
-        if not p.exists():
+        try:
+            stat = p.stat()
+        except OSError:
+            # entry vanished between iterdir() and stat(), or is a broken symlink
             continue
-        stat = p.stat()
         mtime = stat.st_mtime
         fileinfo = {"name": p.name, "modified": mtime}
         if p.is_dir():
-            fileinfo["size"] = len(list(p.glob("*")))
+            # NOTE: not counting directory contents — a glob per subfolder costs one
+            # round-trip each on network filesystems, and the count is not displayed.
+            fileinfo["size"] = 0
             subfolders.append(fileinfo)
         else:
             # files.append(p.resolve().name)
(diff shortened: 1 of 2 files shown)


bmaranville:
One arrangement that seems particularly fragile is a mounted Google drive.  I might try to check this change against that too.

pquarterman17:
I haven't tested with Google drive, so can't comment on that. I do run mine out of Onedrive, and haven't noticed any issues specific to that.

The  navigate up to switch drive fix was something I applied to reductus as well, but I forked so far back with more changes that I'd have to go recreate the fix there to handle multiple drives.

bmaranville:
Is OneDrive the network filesystem that was giving you trouble?  What network filesystems have you tried this with? (we are dealing with NFS most of the time)

pquarterman17:
Onedrive never gave me issues; was more using as a point to compare with Google Drive. I have had issues with changing to network drives, I can't recall if NTS or what, but it was hanging pretty badly.

bumps/webview/server/api.py
```diff
@@ -1498,38 +1499,75 @@ async def get_topic_messages(topic: Optional[TopicNameType] = None, max_num=None
...
+def _get_drives() -> List[str]:
+    """List Windows drive roots, e.g. ["C:\\", "D:\\"]; empty on other platforms."""
+    if hasattr(os, "listdrives"):  # Python 3.12+
+        return os.listdrives()
+    if sys.platform == "win32":
+        import ctypes
+
+        bitmask = ctypes.windll.kernel32.GetLogicalDrives()
+        return [f"{chr(65 + i)}:\\" for i in range(26) if bitmask & (1 << i)]
+    return []
```
pkienzle: I prefer to program to the new interface, with shims marked as cruft to support old interface. In this case put the following at the top and use `listdrives` instead of `_get_drives`:
```
try:
    from os import listdrives
except ImportError:
    # CRUFT: pre-3.12 support
    def listdrives() -> list[str]:
        if sys.platform == "win32":
            ...
        return []
```

pkienzle:
This includes everything #442?

pquarterman17:
Yes, this one is a superset of #442 but given it was a more comprehensive change I put it in as a second PR to make easier to separate out just the network stall aspect of #442

bumps/webview/server/api.py
pquarterman17: Done in 5a8ffd82 — `listdrives` is now imported from `os` at the top with a CRUFT-marked fallback, matching the existing `asyncio.to_thread` shim in the same file, and `_get_drives` is gone. One note: the ImportError branch also covers non-Windows on every Python version (os.listdrives only exists on Windows), where the fallback returns `[]` as before. Verified on Windows/3.13 that the shim's ctypes path produces output identical to the real `os.listdrives`.

bmaranville:
I tested the existing version with the `glob("*")` on our local cluster with NFS, and it was very responsive even with thousands of subdirectory lookups, with < 1s directory listing time.  Your network drive must be _very_ slow! (ours is not usually described as "speedy")

So, it is hard for me to confirm the bug, but I agree that the changes are reasonable and do streamline the process.  I verified that the new implementation works on our NFS system.  Trying out the Windows-specific features next...

bmaranville (approved):
This works as advertised.  Checked on Windows with python == 3.10, and I can use the directory browser.  Checked on Windows with python == 3.12.  Checked on linux with NFS.

PR #437, opened 2026-05-08
Centralize bumps step 4: move support modules out of webview

pkienzle:
Completes #426 refactor

Moves the remaining modules out of bumps/webview/server

webserver.py was moved up one level beside bumps/webview.

We may want to rename it to server.py. I didn't do so in case we wanted to allow for backward compatibility, putting ghost modules in bumps/webview/server that pull their symbols from the new locations. This would allow old code to run mostly without change. (We could improve backward compatibility by moving the main for the new interface out of cli.py, and instead pull in main from old_cli.)

Going forward we should declare the complete public interface, moving most of it to the bumps toplevel namespace, so we are free to reorganize the code without leaving backward compatibility cruft.

Tested using the following script:
```sh
#!/bin/sh

# echo: -x  stop: -e
set -x
set -e

MODEL_NAME=curve
MODEL=doc/examples/curvefit/curve.py
SESSION=/tmp/check.h5
EXPORT=/tmp/check-export
EXPORT2=/tmp/check-export2
rm -rf $SESSION $EXPORT $EXPORT2
# Does the prompt run
bumps --help
bumps --chisq $MODEL
# batch fit new
bumps $MODEL --fit=dream --burn=100 --steps=200 --session=$SESSION --export=$EXPORT --batch
# batch resume
bumps --session=$SESSION --fit=dream --burn=100 --steps=400 --export=$EXPORT2 --batch
ls -l $EXPORT/$MODEL_NAME*.mc.gz $EXPORT2/$MODEL_NAME*.mc.gz
rm -rf $EXPORT2
# webview
bumps $MODEL &

# Maybe make a wxbumps binary for the old version?
bumps --wx --chisq $MODEL
# bumps old gui (wxPython)
bumps --wx --edit $MODEL &
# bumps old cli
bumps --wx $MODEL --resume=$EXPORT --fit=dream --burn=0 --steps=400 --store=$EXPORT2 --batch
ls -l $EXPORT/$MODEL_NAME*.mc.gz $EXPORT2/$MODEL_NAME*.mc.gz

# Test the refl1d plugin
MODEL_NAME=nifilm-fit
MODEL=../refl1d/doc/examples/ex1/nifilm-fit.py
SESSION=/tmp/check-refl1d.h5
EXPORT=/tmp/check-refl1d-export

# webview refl1d
refl1d $MODEL --chisq
refl1d $MODEL &

# refl1d old cli and old gui (wxPython)
python -m refl1d.main $MODEL --chisq
[...]

PR #444, opened 2026-06-06
feat: export fit results summary as <model>-fit.json

### LINKED ISSUE #413: Include χ² in JSON output
Add the overall chi-squared (at full precision) to the -err.json output file. If the error bars are badly over- or under-estimated (chi-squared much different from 1), the confidence intervals sometimes need to be scaled by sqrt(chi-squared). I understand that this is complicated in the multi-curve situation, but I think this should be available so users can make their own determination on how to use it.

Include a column with the scaled uncertainties in the outputs.

Include the degrees of freedom and number of data points so that raw sum of squares can be computed.

pquarterman17:
Closes #413.

*Reworked per review — the original version added chisq/dof/points to the dream `-err.json`; this version instead writes a separate `<model>-fit.json` for all fitters, leaving `-err.json` and `bumps.dream` untouched.*

## What's exported

`export_fit()` now writes `<basename>-fit.json` alongside the other artifacts — a machine-readable summary of the fit results built from the `OptimizeResult`:

| key | value |
|---|---|
| `x`, `dx` | best point and standard error (parallel arrays) |
| `labels` | fitted parameter labels, matching the order of `x`/`dx` |
| `fun` | nllf at the best point |
| `chisq` | normalized χ² at the best point, full precision |
| `dof`, `points` | degrees of freedom, number of data points |
| `method`, `options` | fitter id and options |
| `success`, `status`, `message`, `nit`, `timestamp` | scipy-convention status fields |

The raw sum of squares is recoverable as `chisq · dof`. The fitter state and convergence history are excluded (not JSON serializable / belong with the plottable outputs).

```json
{
  "chisq": 0.775470697827463,
  "dof": 18,
  "dx": [0.0368..., 0.2155...],
  "fun": 6.979236280447168,
  "labels": ["a", "b"],
  "message": "successful termination",
  "method": "amoeba",
  "nit": 44,
  "options": {},
  "points": 20,
  "status": 0,
  "success": true,
  "timestamp": "1781054396.839657",
  "x": [2.5241..., 1.1630...]
}
```

## Design

- New `fitters.fit_result_summary(problem, results)` builds the JSON-serializable dict (whitelisted `OptimizeResult` fields + labels/points/chisq); `fitters.save_fit_result()` writes it. Both live next to `_build_fit_result`.
- `export_fit()` normalizes its `fit` argument: an `OptimizeResult` is used directly (scripting `fit(export=...)` path), a webview `FitResult` goes through `_build_fit_result(problem, fit)` (GUI export and CLI reload paths). One hook covers all fitters and all entry routes.
- `chisq` is computed from `results.fun` via `problem.chisq(nllf=...)`, so the model is not re-evaluated and the value matches `chisq_str` unrounded.
- Note: `dx` comes from the curvature (`problem.cov`) per `_build_fit_result`, so for DREAM it differs from the posterior `std` in `-err.json` — the two files report different (both meaningful) uncertainties.
- `std_scaled` from the previous version is dropped pending the broader decision on scaled χ² handling; with `chisq` and the uncertainties both machine-readable, users can apply `std·√χ²` themselves in the meantime.

## Testing

- `test_fit_result_summary` (fitters.py): summary contents, label/chisq/points math, exclusion of `state`/`convergence`, JSON round-trip — no filesystem needed.
- End-to-end: amoeba fit via `fit(problem, export=...)` and a DREAM fit via `export_fit()` with a webview `FitResult` both produce correct `-fit.json`; the DREAM `-err.json` output is byte-format identical to master.
- pytest on touched modules passes; ruff check/format clean on the diff (no new findings vs master).

### DIFF
--- a/bumps/fitters.py
+++ b/bumps/fitters.py
@@ -1735,6 +1735,9 @@ def _build_fit_result(problem, webview_fit):
         # Non-standard results. Note convergence is in nominal chisq
         dx=dx,
         dof=dof,
+        labels=problem.labels(),
+        numpoints=problem.model_points(),
+        chisq=problem.chisq(nllf=fx),
         # webview FitResults
         state=fit_state,
         convergence=webview_fit.convergence,
@@ -1744,45 +1747,26 @@ def _build_fit_result(problem, webview_fit):
     return results
 
 
-# OptimizeResult entries included in the exported fit results, where available.
-# The fitter state and the convergence history are excluded: the state is not
-# JSON serializable and the convergence history belongs with the plottable
-# outputs rather than the fit summary.
-_FIT_RESULT_EXPORT_FIELDS = (
-    "x",
-    "dx",
-    "fun",
-    "success",
-    "status",
-    "message",
-    "nit",
-    "nfev",
-    "dof",
-    "method",
-    "options",
-    "timestamp",
-)
+# OptimizeResult entries excluded from the exported fit results. The fitter
+# state is not JSON serializable and the convergence history belongs with the
+# plottable outputs rather than the fit summary; both are available in the
+# session file.
+_FIT_RESULT_EXCLUDE_FIELDS = ("state", "convergence")
 
 
-def fit_result_summary(problem, results: OptimizeResult) -> dict:
+def fit_result_summary(results: OptimizeResult) -> dict:
     """
-    Return a JSON-serializable summary of the fit *results* for *problem*.
+    Return a JSON-serializable summary of the fit *results*.
 
-    The summary contains the scipy-convention entries from *results* (x, dx,
-    fun, success, status, message, nit, ...) plus *labels* (the fitted
-    parameter labels, matching the order of x and dx), *points* (number of
-    data points) and *chisq* (normalized chi-squared at the best point, at
-    full precision). For a model with gaussian independent uncertainties the
-    raw sum of squares is chisq*dof.
-
-    The fitter state and the convergence history are not included.
+    The summary contains every entry from *results* except the fitter state
+    and the convergence history. This includes the scipy-convention entries
+    (x, dx, fun, success, status, message, nit, ...) plus *labels* (the
+    fitted parameter labels, matching the order of x and dx), *numpoints*
+    (number of data points) and *chisq* (normalized chi-squared at the best
+    point, at full precision). For a model with gaussian independent
+    uncertainties the raw sum of squares is chisq*dof.
     """
-    summary = {key: results[key] for key in _FIT_RESULT_EXPORT_FIELDS if key in results}
-    summary["labels"] = problem.labels()
-    summary["points"] = problem.model_points()
-    # results.fun is the nllf at the best point; chisq(nllf=None) recomputes it
-    summary["chisq"] = problem.chisq(nllf=results.get("fun"))
-    return summary
+    return {key: value for key, value in results.items() if key not in _FIT_RESULT_EXCLUDE_FIELDS}
 
 
 def save_fit_result(problem, fit, filename):
@@ -1797,50 +1781,56 @@ def save_fit_result(problem, fit, filename):
 
     results = fit if isinstance(fit, OptimizeResult) else _build_fit_result(problem, fit)
     with open(filename, "w") as fid:
-        json.dump(fit_result_summary(problem, results), fid, default=numpy_json, sort_keys=True, indent=2)
+        json.dump(fit_result_summary(results), fid, default=numpy_json, sort_keys=True, indent=2)
 
 
 def test_fit_result_summary():
     class _Problem:
         dof = 8.0
 
+        def getp(self):
+            return np.array([1.0, 2.0])
+
+        def cov(self, x):
+            return np.diag([0.01, 0.04])
+
+        def nllf(self):
+            return 20.0
+
         def labels(self):
             return ["P1", "P2"]
 
         def chisq(self, nllf=None):
-            return 2 * nllf / self.dof
+            return 2 * (self.nllf() if nllf is None else nllf) / self.dof
 
         def model_points(self):
             return 10
 
-    results = OptimizeResult(
-        x=np.array([1.0, 2.0]),
-        dx=np.array([0.1, 0.2]),
-        fun=20.0,
-        success=True,
-        status=0,
-        message="successful termination",
-        nit=100,
-        dof=8.0,
-        method="dream",
-        options={"samples": 100},
-        state=object(),  # not serializable; must be excluded
-        convergence=np.zeros((5, 6)),  # excluded
-    )
-    summary = fit_result_summary(_Problem(), results)
-    assert summary["labels"] == ["P1", "P2"]
-    assert summary["chisq"] == 5.0  # 2*20/8, full precision
-    assert summary["points"] == 10
-    assert summary["dof"] == 8.0
-    assert summary["method"] == "dream"
+    class _WebviewFit:
+        fit_state = None
+        convergence = np.zeros((5, 6))
+        method = "dream"
+        options = {"samples": 100}
+
+    results = _build_fit_result(_Problem(), _WebviewFit())
+    assert results["labels"] == ["P1", "P2"]
+    assert results["chisq"] == 5.0  # 2*20/8, full precision
+    assert results["numpoints"] == 10
+    assert results["dof"] == 8.0
+    assert results["method"] == "dream"
+
+    summary = fit_result_summary(results)
+    # everything except the fitter state and convergence history is exported
     assert "state" not in summary
     assert "convergence" not in summary
+    assert set(summary.keys()) == set(results.keys()) - set(_FIT_RESULT_EXCLUDE_FIELDS)
     # the whole summary must survive a json round trip
     from .dream.stats import numpy_json
 
     data = json.loads(json.dumps(summary, default=numpy_json))
     assert data["x"] == [1.0, 2.0]
     assert data["dx"] == [0.1, 0.2]
+    assert data["chisq"] == 5.0
 
 
 # TODO: Move simplified fit interface to its own file.
@@ -1965,6 +1955,9 @@ def fit(
         # Non-standard results. Note convergence is in nominal chisq
         dx=dx,
         dof=dof,
+        labels=problem.labels(),
+        numpoints=problem.model_points(),
+        chisq=problem.chisq(nllf=fx),
         timestamp=now_string(),
         # webview FitResults
         state=fit_state,
--- a/bumps/fitters.py
+++ b/bumps/fitters.py
@@ -1785,13 +1785,17 @@ def fit_result_summary(problem, results: OptimizeResult) -> dict:
     return summary
 
 
-def save_fit_result(problem, results: OptimizeResult, filename):
+def save_fit_result(problem, fit, filename):
     """
-    Save the :func:`fit_result_summary` of *results* to *filename* as JSON,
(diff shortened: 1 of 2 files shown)


bmaranville:
It is interesting that you noticed that the dream machinery is not directly coupled to `FitProblem` in bumps... this seems like a subtle workaround.  But @pkienzle is the best person to evaluate this.

pkienzle:
TLDR; create OptimizerResult object using `fitters._build_fit_result(problem, state.fitting)` and export that as `*-fit.json`

I would like a common fit results object shared amongst all the fitters which can be saved in the session file or exported as a json file. This should contain the same for a fit driven from the webview interface, from the command line, or using the fit() command from a script or from a notebook. This should be common to all fitters, but with slots for fitter-specific information. Even with the refactor in #437 we aren't there yet.

The state FitResult object currently stores the fitting method, the fit control options such as number of iterations, the convergence curve(s) and the fitter's fitstate. The current program gets parameter labels and values from the model definition. These need to be included in FitResult so that fits can be post-processed without loading the problem definition. Similarly for the parameter uncertainties and/or covariance matrix and degrees of freedom.

The OptimizeResult object returned from the `fit()` call has most of the information we want to save. The field names follow the conventions from scipy optimizers. Because the results are used in scripts and notebooks we should continue use the same names. Keeping the structure of the existing exported json file is less important, though even with these files users may be copying post-processing scripts from fit to fit.

Refactoring state and optimizer is too much for this PR, but we can take steps toward the eventual goal.

We should not add the additional fields to the current DREAM `*-err.json` file. It will be too hard to split the parameter fields from the fit result fields.

For now I suggest writing a json exporter for the OptimizeResult object that can be saved to `*-fit.json`. We should add parameter labels to the result. This will provide fit results for all fitters without requiring any other changes to dream or the export process, or to any of the scripts that load existing `*-err.json` files. It does mean you need to load two JSON files rather than one.

Another option is to add the dream parameter table from `*-err.json` into the `*-fit.json` and no longer write the `*-err.json`. Existing scripts won't break, they just won't work on new fit results.

We may want to include the plottable outputs such as reflectometry curves and profile uncertainties in the session file and in `*-fit.json` but this is a discussion for later.

bumps/dream/stats.py
```diff
@@ -205,17 +205,66 @@ def format_vars(all_vstats):
...
+    gets *std_scaled* = std*sqrt(chisq), the standard deviation scaled for
+    over- or under-estimated measurement uncertainty.
+    """
+    payload = dict((v.label, dict(v.__dict__)) for v in all_vstats)
+    if extra:
+        chisq = extra.get("chisq")
+        if chisq is not None and chisq > 0:
+            scale = np.sqrt(chisq)
+            for entry in payload.values():
+                entry["std_scaled"] = entry["std"] * scale
```
pkienzle: Technically DREAM has no notion of χ² and DOF, only of likelihood. This feels like the wrong place to introduce it.

Scaled standard deviation applies to all fitters. Need to think a little more about how we want to deal with scaled χ² across UI and saved file formats.

bumps/dream/stats.py
```diff
@@ -205,17 +205,66 @@ def format_vars(all_vstats):
...
-            dict((v.label, v.__dict__) for v in all_vstats),
+            payload,
             fid,
             default=numpy_json,
             sort_keys=True,
             indent=2,
         )
 
 
+def test_save_vars(tmp_path):
```
pkienzle: pytest is searching for functions named `test`, `test_*` or `*_test`. It seems this should be failing since it is expecting a path.

For checks outside of CI I call the function `demo*` or `check*` so it doesn't get picked up.

The test file isn't being cleaned up after the test is complete. There may be some helpful context managers in pathlib to manage temporary test file names.

bumps/dream/stats.py
```diff
@@ -205,17 +205,66 @@ def format_vars(all_vstats):
...
+    over- or under-estimated measurement uncertainty.
+    """
+    payload = dict((v.label, dict(v.__dict__)) for v in all_vstats)
+    if extra:
+        chisq = extra.get("chisq")
+        if chisq is not None and chisq > 0:
+            scale = np.sqrt(chisq)
+            for entry in payload.values():
+                entry["std_scaled"] = entry["std"] * scale
+        payload.update(extra)
```
pkienzle: Adding arbitrary fields from `extra` into the dictionary of parameters risks name collision and makes it hard to separate parameters from fields in post-processing.

It would be better to bundle the parameter dict inside extra, but this breaks the existing file format.

Passing extra into the depths of dream is awkward. It may be better to ask dream for the json serializable parameter summary dict and write it outside of the dream package.

bumps/fitters.py
```diff
@@ -860,6 +860,47 @@ def update(k, x, fx, improved):
...
+    (number of data points). For a model with gaussian independent
+    uncertainties the raw sum of squares is chisq*dof.
+    """
+    _, best_logp = state.best()
+    nllf = -best_logp
+    return {
+        "chisq": problem.chisq(nllf=nllf),
+        "dof": problem.dof,
+        "points": problem.model_points(),
+    }
```
pkienzle: This is probably the right place to gather and write the information for the exported fit results json. Maybe not called `fit_quality` though.

If we decide to include the dream parameter summary as part of the fit results json, this should call dream with fit state to build the table.

bumps/dream/stats.py
pquarterman17: I thought pytest applied `tmp_path` which uses a temporary directory and  and then cleans itself up after running? Or maybe I"m misunderstanding. Probably moot point given how much the rest of the PR would need re-working based on other comments.

pquarterman17:
Reworked as suggested — `export_fit` now normalizes to an `OptimizeResult` via `_build_fit_result` and writes `<basename>-fit.json` with parameter labels added; `-err.json` and everything in `bumps.dream` is untouched (the diff no longer touches dream at all). Works for all fitters and all entry routes (webview export, scripting `fit(export=...)`, CLI reload), since they all funnel through `export_fit`.

`std_scaled` is dropped pending the broader decision on scaled χ² — with `chisq` and the uncertainties both exported, users can apply the scaling themselves meanwhile. PR title and description updated to match.

bumps/dream/stats.py
pquarterman17: Resolved by the rework — bumps.dream is no longer touched by this PR; χ²/dof/points now live in the new `-fit.json` written outside dream, and `std_scaled` is dropped for now.

bumps/dream/stats.py
pquarterman17: The reworked test (`test_fit_result_summary` in fitters.py) builds and checks the summary dict directly, including a JSON round-trip, so it no longer touches the filesystem at all —The old `test_save_vars` is gone along with the `save_vars` changes.

bumps/dream/stats.py
pquarterman17: Resolved by the rework — nothing is added to the parameter dict anymore; fit-level and parameter-level values now live in separate files (`-fit.json` / `-err.json`), and the `extra` threading through dream is gone.

bumps/fitters.py
pquarterman17: Replaced by `fit_result_summary()` / `save_fit_result()`, located next to `_build_fit_result`. If the dream parameter table later joins `-fit.json`, `fit_result_summary` is the place to add it.

bumps/dream/stats.py
pkienzle: Sorry... it didn't occur to me that pytest would be looking for the tmp_path parameter name to a test. I should have looked at the docs before commenting.

bumps/webview/server/api.py
```diff
@@ -534,6 +534,13 @@ def export_fit(
     # Write the pars file.
     _write_pars(problem, path, f"{basename}.par")
 
+    # Write a machine-readable summary of the fit results.
+    try:
+        results = fit if isinstance(fit, OptimizeResult) else _build_fit_result(problem, fit)
```
pkienzle: This belongs in the `save_fit_result` function since `_build_fit_result` is nominally private.

bumps/fitters.py
```diff
@@ -1743,6 +1744,101 @@ def _build_fit_result(problem, webview_fit):
...
+    raw sum of squares is chisq*dof.
+
+    The fitter state and the convergence history are not included.
+    """
+    summary = {key: results[key] for key in _FIT_RESULT_EXPORT_FIELDS if key in results}
+    summary["labels"] = problem.labels()
+    summary["points"] = problem.model_points()
+    # results.fun is the nllf at the best point; chisq(nllf=None) recomputes it
+    summary["chisq"] = problem.chisq(nllf=results.get("fun"))
+    return summary
```
pkienzle: I'm neutral on whether *labels*, *chisq*, and *points* should be added.

* *labels* is duplicated elsewhere, but it is convenient to have them when displaying fit results.
* *chisq* is a derived value (`2*fun/dof`), so not really needed.
* *points* isn't used outside of fit problem in bumps; I don't know why you need it.

I suggest adding any extra fields when the result is created in`fit()` and in `_build_fit_result()`. If you need an extra field in the save file then you will need the same field in fit result. That allows you to loop through the fits in a script and process the set of results.

I would blacklist the fields to exclude rather than whitelist the fields to include. This will be less work when the OptimizeResult structure changes.

If we decide to keep the number of points we should call it *npoints* or *numpoints* since it is not the list of *(x,y)* values.

*convergence* is needed to reconstruct an `OptimizeResult` object. It could be written as another field or as a separate data file.

The convergence plot is not exported. Would that be useful?
 
*state* is needed, but it is already exported and is too big to save in json.

*convergence* and *state* are available in a session file, so saving them in export is not that important.

bumps/webview/server/api.py
pquarterman17: Moved into `save_fit_result` — it now accepts either an `OptimizeResult` or a webview `FitResult`.

bumps/fitters.py
pquarterman17: Done — *labels*, *chisq* and *numpoints* are now added when the result is created in `fit()` and `_build_fit_result()`, so in-memory results match the export; `fit_result_summary` now excludes `{state, convergence}` rather than whitelisting fields; and *points* is renamed to *numpoints*.

On the individual fields:

- *chisq*: kept — personally I find it useful to have this exported neatly, as it is a helpful metric, in my opinion.
- *numpoints*: I don't have a specific use in hand; I was erring on the side of providing full context — having *n* alongside *chisq* makes comparisons between models of varied complexity overtly clear. Happy to drop it if you'd rather.
- *convergence*: left out of `-fit.json`. I haven't found a case where I wanted the convergence history after exporting a fit, and it's available in the session file — if someone finds convergence data/plot export useful, that could be follow-on work (it would slot into the existing TODO in `export_fit`).

pkienzle (approved):
Tested and working. Here are is the output from lm on the curvefit example:
```json
{
  "chisq": 1.7553007246376808,
  "dof": 4,
  "dx": [
    0.05798550543486304,
    0.020551405023425583
  ],
  "fun": 3.5106014492753617,
  "labels": [
    "b",
    "m"
  ],
  "message": "successful termination",
  "method": "lm",
  "nit": 0,
  "numpoints": 6,
  "options": {
    "fit": "lm",
    "ftol": 1e-10,
    "jump": 0.0,
    "starts": 1,
    "steps": 200,
    "time": 0.0,
    "xtol": 1e-10
  },
  "status": 0,
  "success": true,
  "x": [
    0.11055072463919124,
    1.9701739130434777
  ]
}
```

Since this involves an output file which will be long-lived, I would like another reviewer to approve the list of exported fields.

PR #445, opened 2026-06-25
API call for setting up a warm start based on current state

hoogerheide:
### Problem statement

Autonomous measurement systems (e.g. AutoRefl, active learning loops) refit a model after every new data point. With DREAM, each cold start discards all information from the previous run and re-draws the initial population from scratch, wasting burn-in time even though the posterior barely moves between iterations.

What is needed is a way for an external controller to:

1. Tell the server that the model has changed (new data injected).
2. Resume DREAM from the chain heads of the previous run rather than a fresh population.

The chain heads are the only information that matters for seeding the next run — the thinned sample history is not needed and becomes a liability (stale log-probability values from the old likelihood surface interfere with best-point tracking on the new problem).

### Requirements

- The controller must be able to replace the `FitProblem` with a new one (same parameters, additional data) without clearing the chain state.
- The degrees of freedom must be recomputed automatically from the new data — no manual update required.
- The payload sent between controller and server must remain small regardless of how many iterations have run or how many samples have been collected. The thinned chain history must not accumulate across iterations.
- If no prior fit state exists (first iteration), the system must fall back gracefully to a cold start.
- If the number of free parameters changes, the system should raise an error — chain heads are meaningless in a different parameter space. Note that in the future there may be use cases where this will be different: for example, adding a model to the fit with its own probe-specific parameters. This can probably be addressed at that point, or we just revert to a cold start.

---

#### New endpoint

**`prepare_warm_start(serialized, method="dataclass", name=None)`**

Replaces the current `FitProblem` (e.g. one with an additional data point) while retaining only the DREAM chain heads from the existing fit state, discarding the thinned sample history, convergence data, and stale log-probability values. The next `start_fit_thread("dream", options, resume=True)` call will seed the chain population from those heads.

The degrees of freedom are recomputed automatically from the new problem, so adding data points requires no manual dof update.

**Fallback behaviour:**
- If no prior fit state exists (first iteration), the chain-head installation is skipped and the subsequent `start_fit_thread` performs a normal cold start.
- If the number of free parameters in `serialized` differs from the current chain state, `ValueError` is raised — chain heads are meaningless in a different parameter space and a cold start via `set_serialized_problem` is required instead.

The `method` parameter carries the same deserialization caveat as `set_serialized_problem`: `"pickle"`, `"cloudpickle"`, and `"dill"` enable arbitrary code execution. The recommended and default method is `"dataclass"`.

#### Intended call sequence per iteration

```python
await sio.call("prepare_warm_start", serialized, "dataclass", name)
await sio.call("start_fit_thread", "dream", options, True)  # resume=True
```

This mirrors the existing cold-start pattern (`set_serialized_problem` + `start_fit_thread`). Fitter options (`samples`, `burn`, `pop`, etc.) are kept separate so they can vary independently of the problem setup.

#### Test

`test_warm_start.py` demonstrates the full loop: a simple Gaussian model with one free parameter (`A`, true value 3.0), adding one noisy data point per iteration. Iteration 1 is a cold start via `set_serialized_problem`; subsequent iterations call `prepare_warm_start` then `start_fit_thread(..., resume=True)`. The estimate of `A` converges toward 3.0 across iterations.

[test_warm_start.py](https://github.com/user-attachments/files/29349292/test_warm_start.py)
[warm_start_output.txt](https://github.com/user-attachments/files/29349293/warm_start_output.txt)

### DIFF
--- a/bumps/dream/core.py
+++ b/bumps/dream/core.py
@@ -299,6 +299,13 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
         # This can happen when a fit is aborted, new data is added to the model, and the fit is resumed.
         # The fit can also be resumed if there are changes to the model code or the constraints as long
         # the list of fit parameters is the same.
+        # Recompute logp for every chain head on the new likelihood surface before
+        # the Metropolis loop starts.  Without this, stale logp values from the
+        # previous fit (which may be much higher or lower under the new data) are
+        # used as the acceptance baseline, causing all new proposals to be either
+        # universally accepted or universally rejected and freezing the chain.
+        logp = dream.model.map(x)
+        state._gen_current_logp[:] = logp
         if state._best_x is not None:
             state._best_logp = float(dream.model.map(state._best_x[None, :])[0])
     else:
--- a/bumps/dream/core.py
+++ b/bumps/dream/core.py
@@ -295,9 +295,10 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
     previous_draws = state.draws
     if previous_draws:
         x, logp = state._last_gen()
-        x = x.copy()
-        logp = dream.model.map(x)
-        state._gen_current_logp[:] = logp
+        # Recalculating the likelihood at the start of resume in case the model definition has changed.
+        # This can happen when a fit is aborted, new data is added to the model, and the fit is resumed.
+        # The fit can also be resumed if there are changes to the model code or the constraints as long
+        # the list of fit parameters is the same.
         if state._best_x is not None:
             state._best_logp = float(dream.model.map(state._best_x[None, :])[0])
     else:
@@ -332,7 +333,7 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
 
     # frame = 0  # For debugging...
     next_outlier_test = max(state.Ngen, 2 * state.Ngen - 10)
-    next_convergence_test = state.generation + state.Ngen if previous_draws else state.Ngen
+    next_convergence_test = state.generation + state.Ngen
     final_gen = dream.draws + dream.burn
     # print(f"running dream with {dream.draws} draws and {dream.burn} burn; final gen={final_gen} draws={state.draws} gen={state.generation}")
     while state.draws < final_gen:
--- a/bumps/dream/core.py
+++ b/bumps/dream/core.py
@@ -295,6 +295,11 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
     previous_draws = state.draws
     if previous_draws:
         x, logp = state._last_gen()
+        x = x.copy()
+        logp = dream.model.map(x)
+        state._gen_current_logp[:] = logp
+        if state._best_x is not None:
+            state._best_logp = float(dream.model.map(state._best_x[None, :])[0])
     else:
         # No initial state, so evaluate initial population
         for x in dream.population:
@@ -327,7 +332,7 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
 
     # frame = 0  # For debugging...
     next_outlier_test = max(state.Ngen, 2 * state.Ngen - 10)
-    next_convergence_test = state.Ngen
+    next_convergence_test = state.generation + state.Ngen if previous_draws else state.Ngen
     final_gen = dream.draws + dream.burn
     # print(f"running dream with {dream.draws} draws and {dream.burn} burn; final gen={final_gen} draws={state.draws} gen={state.generation}")
     while state.draws < final_gen:
--- a/bumps/fitters.py
+++ b/bumps/fitters.py
@@ -886,51 +886,30 @@ class DreamModel:
         return -np.array(self.mapper(pop))
 
 
-def _rebuild_mcmc_state(old_fit_state, fitProblem, options: dict):
-    """Build a minimal MCMCDraw for a parameter-space change (add/remove/reorder).
+def map_labels(labels: list, old_labels: list) -> list:
+    """Return (new_index, old_index) pairs for every label in *labels* that also
+    appears in *old_labels*.
 
-    Uses ``initpop.generate`` with the stored fit options to produce a
-    correctly-sized population for the new problem (same pop scaling and init
-    strategy as DreamFit.solve).  Matched parameters have their columns
-    overwritten with old chain-head values; new parameters keep generated values.
+    Duplicates are matched in order: the first occurrence of a value in *labels*
+    maps to the first unused occurrence in *old_labels*, the second to the second,
+    and so on.
 
-    Returns a new MCMCDraw with ``draws=new_Npop`` so that ``_run_dream`` enters
-    the resume path, triggering nllf recompute and convergence delay.
+    # TODO: Use map_labels() to simplify fitproblem.load_pars()
     """
-    from .dream.state import MCMCDraw
+    from collections import defaultdict
 
-    old_labels = old_fit_state.labels
-    new_labels = fitProblem.labels()
-    old_x, _ = old_fit_state._last_gen()  # (old_Npop, old_Nvar)
+    # Build a list of old positions for each label (to handle duplicates).
+    idx_map: defaultdict = defaultdict(list)
+    for j, s in enumerate(old_labels):
+        idx_map[s].append(j)
 
-    # generate() handles pop scaling, init strategy, bounds, and use_point.
-    new_pop = initpop.generate(fitProblem, **options)  # (new_Npop, new_Nvar)
-    new_Npop, new_Nvar = new_pop.shape
+    pairs = []
+    for i, s in enumerate(labels):
+        if idx_map[s]:
+            j = idx_map[s].pop(0)
+            pairs.append((i, j))
 
-    # Override columns for parameters that survived from the old chain.
-    # np.resize tiles old rows to fill new_Npop.  Duplicated rows diverge quickly
-    # via DE perturbations; this is preferable to cold random draws from generate().
-    old_label_to_col = {label: i for i, label in enumerate(old_labels)}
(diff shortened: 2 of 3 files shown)


hoogerheide:
Another option which would also work with the CLI is to adopt what I did in `AutoRefl` where I made a new DreamFit class:

```
class DreamFitPlus(DreamFit):
    def __init__(self, problem):
        super().__init__(problem)

    def solve(self, monitors=None, abort_test=None, mapper=None, initial_population=None, **options):
        ...
        population = initpop.generate(self.dream_model.problem, **options) if initial_population is None else initial_population

```

The way this would work would be to include an `initial_population` key in `options`; fall back to the `init` method if `None` otherwise use the provided population. Then the client just passes this through.

It would require some guards on the GUI (always set `initial_population=None` if using the Start button in the GUI, for example, and make sure there's no text entry field for it.

pkienzle:
I think we are better of using the existing load and resume infrastructure. With only one code path there is less chance of missing something like the crossover ratio settings.

If we recompute the nllf on the chain heads and on state.best at the start of resume that covers most of the problem. Even the convergence plot will update itself smoothly.

The convergence test will still be an issue. After changing nllf we should resume with `next_convergence_test = state.generation + state.Ngen` rather than `next_convergence_test = state.Ngen`. This will force it to clear out the existing buffer before testing for triggering convergence tests after every batch of steps (currently `DE_steps=10`).

It should be okay to do both of these on every resume. The initial calculation is an additional work step, but that is small compared to the total DREAM run. Delayed convergence testing is also not a problem. With auto trimming to move the burn point within the buffer we only need resume when we want to throw away most of it away.

This strategy will also work for resume after changing the priors, burning through the entire buffer before checking convergence.

hoogerheide:
@pkienzle Is this what you had in mind? I hesitated to touch `core.py` but this seems to work. Note that the warm start API call is now just a replace_serialized_problem, which is just set_serialized_problem without the state reset. The state reset could be a flag on set_serialized_problem if you prefer.

bumps/dream/core.py
```diff
@@ -295,6 +295,11 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
     previous_draws = state.draws
     if previous_draws:
         x, logp = state._last_gen()
+        x = x.copy()
```
pkienzle: Copy not needed here.

You don't need to update `_gen_current_logp` here. It'll happen automatically in the call to `self._generation()` which completes the first update step. I suppose it is safer to do so even if it is not currently used.

`state._best_logp` does need to be updated because the current value is checked against the new logp values during `self._generation()`.

Add a comment to say why we are recalculating logp:
```
# Recalculating the likelihood at the start of resume in case the model definition has changed.
# This can happen when a fit is aborted, new data is added to the model, and the fit is resumed.
# The fit can also be resumed if there are changes to the model code or the constraints as long
# the list of fit parameters is the same.
```

bumps/dream/core.py
```diff
@@ -327,7 +332,7 @@ def _run_dream(dream: Dream, abort_test=lambda: False):
 
     # frame = 0  # For debugging...
     next_outlier_test = max(state.Ngen, 2 * state.Ngen - 10)
-    next_convergence_test = state.Ngen
+    next_convergence_test = state.generation + state.Ngen if previous_draws else state.Ngen
```
pkienzle: Drop the conditional. If there are no previous draws then state.generation will be zero.

bumps/webview/server/api.py
```diff
@@ -426,6 +426,58 @@ async def load_session(pathlist: List[str], filename: str, read_only: bool = Fal
...
+    ``"pickle"``, ``"cloudpickle"``, or ``"dill"``.  The latter three enable
+    arbitrary code execution and carry the same caveat as
+    ``set_serialized_problem``.
+    """
+    # Deserialize the new problem first so we fail fast on bad input.
+    fitProblem = deserialize_problem(serialized, method=method)
+
+    # Validate parameter count against the current chain state.
+    fit_state = state.fitting.fit_state
+    if fit_state is not None and hasattr(fit_state, "Nvar"):
```
pkienzle: You could instead check `state.labels == fitProblem.labels()`. This is robust against adding, subtracting and reordering parameters.

bumps/webview/server/api.py
```diff
@@ -426,6 +426,58 @@ async def load_session(pathlist: List[str], filename: str, read_only: bool = Fal
     state.shared.updated_parameters = now_string()
 
 
+@register
+async def replace_serialized_problem(
```
pkienzle: I would call it `update_serialized_problem` since that sounds a little less drastic than replacing it. YMMV

pkienzle (approved):

hoogerheide:
These changes are made here.

hoogerheide:
A couple of things: https://github.com/bumps/bumps/pull/445/commits/4a57faa5bb73c454ef7df63a6887457cfb8384e4 restores the logp calculation in core.py; this turns out to be essential so chains don't freeze by keeping old logp values that might be significantly better than new ones (due to surprise from new data points) and so rejecting all metropolis steps.

hoogerheide:
Also, it turns out it's not that hard to support added or removed parameters in a warm start environment. https://github.com/bumps/bumps/pull/445/commits/a1ed3e0c4554e3122950eea203ec52f0c79bc88d does with the API only; https://github.com/bumps/bumps/pull/445/commits/095fc22b5415336aff81ddeb7f98207be8055ed5 does this internally in the `DreamFit.solve` logic.

Here's an updated test script, including tests for adding and removing parameters:
[test_warm_start.py](https://github.com/user-attachments/files/29385803/test_warm_start.py)

hoogerheide:
Added parameters are particularly important when building up linked models using autonomous data collection techniques (e.g. D2O contrast, then start collecting H2O data but warm start using the existing D2O data).

bmaranville:
How are you keeping track of parameters?  By index?

hoogerheide:
> How are you keeping track of parameters? By index?

It's by label.

```
    # Override columns for parameters that survived from the old chain.
    # np.resize tiles old rows to fill new_Npop.  Duplicated rows diverge quickly
    # via DE perturbations; this is preferable to cold random draws from generate().
    old_label_to_col = {label: i for i, label in enumerate(old_labels)}
    old_x_resized = np.resize(old_x, (new_Npop, old_x.shape[1]))
    for new_col, label in enumerate(new_labels):
        if label in old_label_to_col:
            new_pop[:, new_col] = old_x_resized[:, old_label_to_col[label]]
```

bumps/fitters.py
```diff
@@ -924,6 +971,14 @@ def update(state, pop, logp):
...
             steps = (draws + pop_size - 1) // pop_size
         monitors.info(f"# burn: {options['burn']} # steps: {steps}, # draws: {pop_size * steps}")
+
+        # On resume, if the parameter space changed (add/remove/reorder) rebuild the
+        # chain state so _run_dream enters the resume path (nllf recompute + convergence
+        # delay) rather than starting cold.
+        if self.state is not None and hasattr(self.state, "labels"):
+            if self.state.labels != self.problem.labels():
+                self.state = _rebuild_mcmc_state(self.state, self.problem, options)
+
```
pkienzle: Excellent idea. Now we can resume arbitrary models even after changing the fit parameters.

Rather than modifying state when the labels have changed, just update the initial population from the preserved labels and pretend that there is no state. So something like (untested):
```python
from collections import defaultdict

...

if self.state:
    labels, old_labels = self.problem.labels(), self.state.labels
    if labels != old_labels:
        new_index, old_index = zip(*map_labels(labels, old_labels))
        old_population, _ = self.state._last_gen()
        population[new_index] = old_population[old_index]
        self.state = None

...

# TODO: Use the map_labels() function to simplify fitproblem.load_pars()
def map_labels(labels: list[str], old_labels: list[str]) -> list[tuple[int, int]]:
    """
    Return a list of (new_index, old_index) pairs for every element in ``labels``
    that also appears in ``old_labels``.  Duplicates are matched in order:
    the first occurrence of a value in ``labels`` maps to the first unused
    occurrence in ``old_labels``, the second to the second, etc.
    """
    # Collect all positions of each string in old_labels. To handle duplicates, each
    # label has a list of indices.
    idx_map: defaultdict[str, list[int]] = defaultdict(list)
    for j, s in enumerate(old_labels):
        idx_map[s].append(j)

    # Walk new labels, consuming the earliest unused old index
    pairs: list[tuple[int, int]] = []
    for i, s in enumerate(labels):
        if idx_map[s]:                     # there is at least one unused match
            j = idx_map[s].pop(0)          # take the first (earliest) old index
            pairs.append((i, j))

    return pairs
```

Users have asked for the ability to zap an existing parameter and reinitialize it from LHS because they think it is stuck in a bad minimum. I think we could handle this with an additional zap labels option passed into map_labels, so the above proposal is good enough for now. The zap labels option would apply for all fitters when --resume or --pars is given on the command line, so too much to squeeze into this PR.

pkienzle: Need another tweak: when resuming we can't change the number of chains.

Something like:
```python
if self.state:
    # Number of chains can't change on resume, so ignore pop option and use existing value.
    population = initpop.generate(self.problem, init=options["init"], pop=-self.state.Npop) 
    labels, old_labels = self.problem.labels(), self.state.labels
    if labels != old_labels:
        new_index, old_index = zip(*map_labels(labels, old_labels))
        old_population, _ = self.state._last_gen()
        population[new_index] = old_population[old_index]
        self.state = None
else:
    population = initpop.generate(self.problem, init=options["init"], pop=options["pop"])
population = population[None, :, :]
```

hoogerheide:
@pkienzle Good suggestions and somewhat simpler. They are implemented here. I have tested against the latest version of the test_warm_start script and everything looks good.

pkienzle (approved):

PR #456, opened 2026-07-22
ACNS tuturial notebooks

pkienzle:
Tutorial notebooks used for the ACNS 2026 tutorial on Monte Carlo uncertainty propagation and bumps fitting.

Also updates the installation docs.

Based off changes from #447, the fix-jupyter-startup branch.

You can test the notebooks on google colab.

* Select > File > Open notebook in the menu
* Select > GitHub from the data sources
* Set search box to "bumps"
* Select the "bumps/bumps" repository
* Select the "acns-notebooks" branch

Three notebooks will appear. You should be able to select any of them and click on "Run all". This will install the package and run the cells through completion.

I don't check that the notebooks run to completion during CI. AI suggestions papermill:
```yaml
- name: Test Tutorials
  run: |
    pip install papermill ipykernel
    # Loop through all notebooks and execute them
    for nb in tutorials/*.ipynb; do 
      papermill "$nb" "executed_${nb}" || exit 1
    done
```
For now I created a `check_notebooks.sh` which does the loop, but is too slow to run on every commit (5 min to run the latex_sphere.ipynb examples)


bumps/webview/webserver.py
```diff
@@ -397,7 +397,7 @@ async def start_app(
     else:
         print(f"webserver started: {get_server_url()}")
 
-    return api
+    return None
```
bmaranville, line 400: This is one of the only things giving me pause... Why change the return type and value?

bumps/webview/webserver.py
```diff
@@ -434,7 +434,26 @@ def display_bumps(width: Union[str, int] = "100%", height: Union[str, int] = 120
...
     :param width: The width of the iframe.
     :param height: The height of the iframe.
+    :param single_panel: The title of the panel to display.
+
+    Note that the single panel mode shows only that panel. It does not show the menu bar or
+    the panel tabs, so the panel cannot be changed.
     """
+    try:
+        # If running in a colab notebook then use the google colab proxy to display webview
+        from google.colab import output
```
bmaranville, line 444: It seems strange to import a Google specific library when most jupyter use will not be in colab

bumps/webview/webserver.py
pkienzle: Because `<module 'bumps.api' from '.../bumps/api.py'>` in the cell output is ugly... perhaps not the best of reasons.

Also, I've been linking the bits of api that are useful for scripts and jupyter in the the bumps.names namespace, so there should be very little reason to access api directly.

bumps/webview/webserver.py
pkienzle: A primary use case for colab is tutorial sessions. I want to minimize friction getting started.

If the code were shorter I would put it directly in the notebook.

Also, putting it in bumps means that I can fix it in one place when colab changes. Otherwise I would have to update every tutorial when they change the startup process.

doc/notebooks/ACNS-2026-GUM.ipynb
```diff
@@ -0,0 +1,1114 @@
...
+    "\n",
+    "    # std uses the bias correction\n",
+    "    std = np.sqrt(np.sum((samples - mean) ** 2) / (N - 1))  # unbiased\n",
+    "\n",
+    "    # Sort the samples to get the quantiles\n",
+    "    index = np.argsort(samples)\n",
+    "    samples = samples[index]\n",
+    "    if logp is not None:\n",
+    "        logp = logp[index]\n",
+    "    median = samples[N // 2]  # could average central pair if N is odd\n",
```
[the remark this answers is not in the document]
pkienzle: Don't want `np.median` here because the samples are already sorted.

The even/odd condition is indeed reversed: could average central pair if N is even

bumps/fitters.py
```diff
@@ -1524,7 +1524,7 @@ def help(*shown):
         webview="""
 Webview functions:
 ```python
-await bp.start_bumps(session="results.h5", ...)      # start the webview server; see help("startup") for options
+await bp.start_bumps()                               # start the webview server; see help("startup") for options
```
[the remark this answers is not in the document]
pkienzle: The usual case is to start the server without a session. Don't complicate the cheat sheet with the optional parameter.

